PR #28, opened 2016-03-17
Ignore generated .so files

embray:


jdemeyer:
Which generated `.so` files do you see? They should only be in the `build` directory, which is already ignored.

embray:
When using `./setup.py build_ext --inplace` they are placed in the src directory.  This is convenient for testing.

PR #29, opened 2016-03-22
Some proposed setup.py changes

embray:
This PR makes a few changes to the setup.py to make it a bit cleaner, IMO.  It ultimately made the most sense to make all these changes simultaneously, but I can try to break out the individual changes more cleanly if desired.
- The main change is that most code from the previous setup.py the generated files--either running `./configure` or `cythonize`--are encapsulated as subroutines, and those subroutines are all run as methods on a customized `build_ext` command.  This ensures that the `./setup.py` script can be run without repeating those steps every time it is run, and instead those steps are only performed when it is relevant to (i.e. when building).  This is the most important change here.
- The files generated by `./configure` are not placed in `build/`, since the directory used by distutils for builds is not necessarily hard-coded.  An alternative would be to default to `build/`, but still accept a different output directory as an option to the configure script.  Personally I find it easier to understand when the generated files are still output alongside the source code rather than in a parallel tree, but that's arguable.
- Similarly, files generated by Cython are kept alongside their source files--again, I find this easier to manage, and it doesn't require any tricks with `package_data`.
- In order to prevent generated files from cluttering the git status, and to make it easy to clean the src/ directory, a `.gitignore` for all generated files is auto-generated and placed in src/.


jdemeyer:
Personally, I really prefer a clean separation between source files and build files. For me, it's a feature that all build artifacts are in `build`.

jdemeyer:
> since the directory used by distutils for builds is not necessarily hard-coded.

I don't really get this argument. Nothing is hard-coded, it's explicit in `setup.py`.

embray:
The path `build/` is not hard-coded.  If one runs `./setup.py build -b temp` the build artifacts are placed in `temp/` instead of `build/`.

Now, I just tested the setup.py under master with this, and it does still work because all the Cython output is currently hard-coded to go into `build/`.  And it's rare that anyone will use this option, but it's still best to to not assume.  Also confusing is that under `build/` some files are going into `build/cysignals` while others are going into `build/src/cysignals`.

For Cython I find it easier to deal with if the generated files are created alongside their associated pyx and pyd files.  For one it makes debugging a bit easier--it's not uncommon to have to look at the .c files to see what Cython is outputting, and it's annoying if they're placed in a parallel but separate "source" tree.  It's worse for generated header files, because now you have headers coming from multiple directories instead of all being in one place.  Finally, it makes it easier (though this is more a limitation of distutils than a real technical barrier) to include the generated C sources in the source distribution.  This makes it easier to distribute the package without requiring the person installing to have Cython (or a new enough Cython for that matter).

I sympathize with preferring to keep build artifacts completely segregated and in other cases I'd agree, but I've found that dealing with Cython is much easier if that's not the case.  I've yet to see a good argument for how the end result is different in any practical sense, especially if the generated files are consistently ignored by git.

embray:
Oh, debugging Cython code in gdb (as described [here](http://docs.cython.org/src/userguide/debugging.html)) tends to work a little better in this case too, though it can technically work either way I think).

jdemeyer:
> Also confusing is that under build/ some files are going into build/cysignals while others are going into build/src/cysignals.

FYI: this is a [Cython bug](https://github.com/cython/cython/pull/486) fixed in the upcoming version 0.24.

embray:
Ah, I figured that was the case--thanks for fixing it.  That was annoying.

jdemeyer:
Whether or not one prefers a separate build directory is a personal preference, one can give arguments both ways. It happens that my personal preference is to keep things separated.

Do you know of a solution which could support both ways?

embray:
I thought you might object to that--it's really not the most important part of this PR.  If you feel strongly about it I can revert that change while keeping the main part which is better encapsulation of the individual build steps.

jdemeyer:
First of all, I don't think there is a problem with `setup.py`: it does what it needs to do, it works. I certainly agree that it's ugly. But that's mainly for reasons that I cannot control, for example the `cython` issue and the fact that `package_data` cannot be installed from arbitrary directories.

That being said, if you want to clean up `setup.py` without changes in functionality, go ahead. I don't really care how `setup.py` looks like as long as it works as well as before.

setup.py
```diff
@@ -1,9 +1,15 @@
...
-from distutils.command.build_py import build_py
-from distutils.extension import Extension
-from Cython.Build import cythonize
+try:
+    from setuptools.command.build_ext import build_ext as build_ext
+    from setuptools import setup, Distribution, Extension
+except ImportError:
+    from distutils.command.build_ext import build_ext as build_ext
+    from distutils.core import setup, Distribution
+    from distutils.extension import Extension
```
jdemeyer: I really don't like these conditional imports. Either use `setuptools` or use `distutils` but I don't want to support both.

embray: Fine by me.

embray:
> First of all, I don't think there is a problem with setup.py: it does what it needs to do, it works. 

Technically it is broken, albeit subtly.  This goes back to one of the many original sins of distutils' design, in this case the sin inherent in allowing a totally free-form setup.py script.  The problem is that a setup.py script is often assumed to have many undocumented, or at least poorly documented invariants that must be kept.  Perhaps the most simple is that running

```
./setup.py --version
```

should just print the package version, and never anything else.  So in general running the `./setup.py` script should not have any *side-effects* except in the specific cases where those side-effects are expected.  For example generating build output files should only occur in a build-related command and never otherwise.  There are of course exceptions to this but those exceptions should be handled carefully.

embray:
> I don't really care how setup.py looks like as long as it works as well as before.

It didn't work well for me before.

jdemeyer:
> It didn't work well for me before.

I guess I meant: don't break my workflow and don't break Travis CI :-)

embray:
Well obviously I'll fix whatever the problem is on Travis.

jdemeyer:
Another issue: the file `__init__.pxd` really should be generated in the install phase, since `setup.py build` might be called without `--prefix` but the path which is written in `__init__.pxd` depends on the `--prefix`.

embray:
That makes some sense--`__init__.pxd` is only needed when installing anyways.

jdemeyer:
The build system has changed significantly since this pull request and some issues have in fact been fixed in the mean time. I'm closing this, feel free to open a new PR if further changes are needed.

PR #36, opened 2016-09-08
Enhance helper handling in cysignals-CSI

jgmbenoit:
This patch avoid to install the helper Python include script in the PREFIX/bin directory, given that installing non executable files in this directory is highly not recommended. Note that the extension of the helper Python include script was changed from .py to .pii : the rational behind is to avoid  it to be byte-compiled by setup.py and alike scripts.


jdemeyer:
> the extension of the helper Python include script was changed from .py to .pii

This is silly. It's Python code, so it should have a `.py` extension.

jgmbenoit:
Concerning the extension, I was looking for a mean to avoid byte-compiling by the scripts. I could not figure out a better way. I am agree to say that the helper contains Python code, but as you claimed in a previous discussion, it is read as a data file by the cysignals-CSI script: so, it not really a Python script from this perspective, it just contains Python code that is literally read by an other Python script and keeping a .py extension misleads automate scripts and potential human readers.

jdemeyer:
I don't think it is such a problem that the script is byte-compiled. It's a bug in the tool, but I can live with it.

jgmbenoit:
The problem is rather the presence of the byte-compiled output which is clearly not used.
Anyway, the main issue is not the extension of the helper, but its presence in the PREFIX/bin directory, so reverting .pii to .py is not the main subject here.

jgmbenoit:
May I revert the extension ?

infinity0:
Some other methods for avoiding byte-compiled code is here: https://stackoverflow.com/questions/154443/how-to-avoid-pyc-files

jgmbenoit:
Thanks @infinity . It appears that it would be better to put the gdb helper not in a subfolder but in a folder of its own as it is architecture independent contrary to the core part. Fortunately, this may also solve the  extension issue.

jgmbenoit:
https://github.com/sagemath/cysignals/pull/39

PR #39, opened 2016-09-15
Enhance helper handling in cysignals-CSI

jgmbenoit:
This patch avoid to install the helper Python include script in the
/usr/bin directory, given that installing non executable files in
this directory is highly not recommended.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -130,11 +130,13 @@ setup(
     platforms=["POSIX"],
 
     ext_modules=extensions,
-    packages=["cysignals"],
+    packages=["cysignals","cysignals_gdb"],
     package_dir={"cysignals": opj("src", "cysignals"),
-                 "cysignals-cython": opj(cythonize_dir, "src", "cysignals")},
+                 "cysignals-cython": opj(cythonize_dir, "src", "cysignals"),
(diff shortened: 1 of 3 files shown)


jdemeyer:
Instead of complicating the `setup.py` file, it would be better to reorganize the sources as well.

In any case, the fact that you needed to add the `__init__.py` is bad, since `src/scripts` is **not** a Python package.

jgmbenoit:
I am agree to say that the patch is not ideal. But at least no non-executable file is installed in PATH.
In UN*X system the involved script can find its place in  /usr/share/cysignals/ , but this approach is UN*X specific. The best remain to put the content of the involved in the calling script itself.

jdemeyer:
This patch makes the build system more complicated than necessary. That should be fixed.

jgmbenoit:
May be, but the the initial issue remains meanwhile.

jdemeyer:
I'm going to close this for now. If you want to continue the discussion, please open 1 pull request for each issue.

jdemeyer:
A made a different and better fix in master. Now the helper is installed in `share/cysignals`.

PR #40, opened 2016-10-08
Soften docs/Makfile to avoid clean to stop

jgmbenoit:
Description: upstream: documentation composition: soften user-friendly check
 Render `user-friendly check for sphinx-build' even more friendly by not
 emitting the error message and not stopping wherever is not needed in view
 to be able to launch non-composition targets (e.g., clean) on systems without
 sphinx-build installed; meant to be submitted to the upstream maintainer.
Origin: vendor, Debian
Comment: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=840017
Author: Jerome Benoit <calculus@rezozer.net>
Last-Update: 2016-10-08

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -7,11 +7,6 @@ SPHINXBUILD   = sphinx-build
 PAPER         =
 BUILDDIR      = build
 
-# User-friendly check for sphinx-build
-ifeq ($(shell which $(SPHINXBUILD) >/dev/null 2>&1; echo $$?), 1)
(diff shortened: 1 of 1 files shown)


jdemeyer:
This `Makefile` is auto-generated. If you don't like it, you should file a bug with [Sphinx](https://github.com/sphinx-doc/sphinx)

jgmbenoit:
Autogenerated as automake does or as template ? I will check. Thanks for the link and sorry for the noise.

PR #50, opened 2017-03-30
Forward declare symbols to avoid need for late-include.

robertwb:


jdemeyer:
Thanks. This doesn't quite work yet, follow-up at #51.

PR #51, opened 2017-03-30
Forward declarations

jdemeyer:
This continues and fixes #49 and #50 in order to avoid a need for `.pxi` files.


jdemeyer:
It's uglier than #49 because of this
```C
#ifdef CYSIGNALS_IMPLEMENTATION
static cysigs_t cysigs;
static void _sig_on_interrupt_received(void);
static void _sig_on_recover(void);
static void _sig_off_warning(const char*, int);
#else
static cysigs_t *__pyx_vp_9cysignals_7signals_cysigs;
#define cysigs (*__pyx_vp_9cysignals_7signals_cysigs)
static void (*_sig_on_interrupt_received)(void);
static void (*_sig_on_recover)(void);
static void (*_sig_off_warning)(const char*, int);
#endif
```
but it seems to work with a stock Cython.

jdemeyer:
Unfortunately, this doesn't work in C++ because
```
static int x;
static int x = 0;
```
is not legal C++ code. Apparently, you cannot forward-declare variables in C++: http://stackoverflow.com/questions/3714808/forward-declarations-for-variables

robertwb:
That's unfortunate (both C++ and the function-pointer vs. not inconsistency). But this might work: define an inline function `get_cysigs` in cysignals.pxd that returns a pointer to the cysigs_t variable which *can* be forward declared, and define cysigs to be that dereferenced. 

You could also use this trick (inline functions declared in the pxd) to avoid the CYSIGNALS_IMPLEMENTATION toggling (where the body would simply call the non-inlined variant).

jdemeyer:
This now finally seems to work. I hope you agree that it's not a clean solution and that something like https://github.com/cython/cython/pull/1650 is still useful.

PR #52, opened 2017-03-30
WIP Fix gdb debugging and test it on Travis CI

jdemeyer:
The `cysignals-CSI` is currently broken for different reasons:
- [x] The Python library which is being debugged and the Python runtime started by GDB may mismatch.
- [x] There are various `bytes` vs. `str` incompatibilities in the script.
- [ ] Cython debugging does not work with Python 3.4 due to https://github.com/cython/cython/issues/1655

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,20 +8,25 @@ python:
 
 before_install:
   - sudo apt-get update -qq
-  - sudo apt-get install -qq -y python-pip python-dev
-  - sudo apt-get install -qq -y python-flake8
+  - sudo apt-get install -q -y python-pip python-dev
+  - sudo apt-get install -q -y python-flake8
+  - sudo apt-get install -q -y gdb python2.7-dbg python3.4-dbg
 
 install:
   - pip install -r requirements.txt
   - cython --version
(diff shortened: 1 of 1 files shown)


anntzer:
Also, would it be possible to have a way to just disable gdb debugging?  I mostly just want to use cysignals', well, signal handling facility, and only work on Py3, so at that point the tracebacks generated by gdb failures are just noise to me.

jdemeyer:
> Also, would it be possible to have a way to just disable gdb debugging?

What do you have in mind? A compile-time setting, run-time setting, environment variable...?

anntzer:
I think gdb integration should be turned on either by an environment variable (my preference, as it is the easiest to set without modifying the source code) or by a runtime function (e.g. `cysignals.run_gdb_on_failure()`).  Or possibly even by a C macro, e.g. NDEBUG. (not sure I really like that idea).

collares:
testgdb.py says it "is known to be broken in some cases, see [this issue]", but this issue was closed. Does this mean previously-known issues were fixed, or just are no longer relevant?

PR #54, opened 2017-04-21
Squelch some debug messages when debug_level = 0

embray:
Protect debugging message with a `cysigs.debug_level` check.  This message breaks one or two tests in Sage when built with debug enabled.  I don't know what debug_level it should be at but at least 1, I assume.

### DIFF
--- a/src/cysignals/macros.h
+++ b/src/cysignals/macros.h
@@ -239,7 +239,7 @@ static inline int sig_check(void)
 static inline void sig_block(void)
 {
 #if ENABLE_DEBUG_CYSIGNALS
-    if (cysigs.block_sigint != 0)
+    if (cysigs.debug_level >= 1 && cysigs.block_sigint != 0)
     {
         fprintf(stderr, "\n*** WARNING *** sig_block() with sig_on_count = %i, block_sigint = %i\n", cysigs.sig_on_count, cysigs.block_sigint);
         print_backtrace();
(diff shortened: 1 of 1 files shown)


jdemeyer:
> This message breaks one or two tests in Sage

Those are genuine bugs.

Your patch is shooting the messenger instead of accepting the message.

embray:
If that's a genuine error then should that code be bracketed with `#ENABLE_DEBUG_CYSIGNALS` in the first place then?  Because without debug I don't get the message at all and the code in question (seems to) work fine.  So which is it?

embray:
Right, so I think it's reasonable to be able to disable it.  [Here](https://github.com/sagemath/cysignals/blob/master/src/cysignals/implementation.c#L274) you have `_sig_off_warning` which doesn't check `cysigs.debug_level` but is always compiled regardless of `#ENABLE_DEBUG_CYSIGNALS`.  Obviously what it's warning about is a programming error that should be warned about, but it seems inconsistent...

jdemeyer:
The reason for the difference is that `sig_off()` is much more common than `sig_block()`. So the chance that somebody misuses `sig_off()` is much larger than the chance that somebody misuses `sig_block()` (just because the chance that person P *misuses* X is bounded by the chance that P *uses* X).

embray:
Fair enough. I just though it was surprising that I would get this warning with debug on, but not with debug off.

PR #57, opened 2017-05-12
Increase FD_SETSIZE from its default of 64 on Cygwin

embray:
This is basically the same problem that [tso-22839](https://trac.sagemath.org/ticket/22839) is addressing, but applied to Cysignals (specifically the pselect module).  It's necessarily to manually bump up `FD_SETSIZE` to something reasonable, and ISTM the easiest way to do that consistently, especially with Cython, is to pass it at build time.

This came up in Sage when running `./sage -t` with a large number of processes.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,6 +30,9 @@ cythonize_dir = "build"
 macros = []
 
 if sys.platform == 'cygwin':
+    # On Cygwin FD_SETSIZE defaults to a rather low 64; we set it higher
+    # for use with PSelecter
+    # See https://github.com/sagemath/cysignals/pull/57
     macros.append(('FD_SETSIZE', 512))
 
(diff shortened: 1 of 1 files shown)


jdemeyer:
Please add some documentation (the explanation in this PR could be a comment in the code).

embray:
Ok, no problem--I figured you'd ask. Just wanted to make sure you were OK with the general approach first.

embray:
Thanks--how soon do you think we can get this update into Sage? It's not super important but nice to have for running the tests.

jdemeyer:
> how soon do you think we can get this update into Sage?

5 minutes if needs be :-)

embray:
Well, okay. It's not an emergency, but thanks!

jdemeyer:
See https://trac.sagemath.org/ticket/23537

PR #58, opened 2017-05-20
Remove mucking with environment variables

vbraun:
Do not change:
* LD_PRELOAD
* DYLD_INSERT_LIBRARIES
* PYTHONPATH

### DIFF
--- a/src/scripts/cysignals-CSI
+++ b/src/scripts/cysignals-CSI
@@ -101,9 +101,7 @@ def run_gdb(pid, color):
 
     env = dict(os.environ)
     try:
-        # Run gdb but pretending to be Python in argv[0].
-        # This is needed for Python to find its own files.
-        cmd = Popen([sys.executable], executable=whichgdb,
+        cmd = Popen(["gdb"], executable=whichgdb,
                 stdin=PIPE, stdout=PIPE, stderr=PIPE, env=env)
     except OSError:
         return b"Unable to start gdb (not installed?)"
--- a/src/scripts/cysignals-CSI
+++ b/src/scripts/cysignals-CSI
@@ -95,23 +95,11 @@ def run_gdb(pid, color):
     """
     Execute gdb.
     """
-    env = dict(os.environ)
-
-    # Preload the right Python library
-    libpython = os.path.join(sysconfig.get_config_var('LIBDIR'),
-                             sysconfig.get_config_var('INSTSONAME'))
-    if sys.platform == 'macosx':
-        env['DYLD_INSERT_LIBRARIES'] = libpython
-    else:
-        env['LD_PRELOAD'] = libpython
-
-    # Set the correct Python path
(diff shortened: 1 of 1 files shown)


jdemeyer:
On my system, this makes things worse unfortunately...

Within Sage, I get
```
  File "<string>", line 26, in <module>
  File "/usr/local/src/sage-config/local/lib/python2.7/site-packages/Cython/Debugger/libpython.py", line 61, in <module>
    import tempfile
  File "/usr/local/src/sage-config/local/lib64/python2.7/tempfile.py", line 32, in <module>
    import io as _io
  File "/usr/local/src/sage-config/local/lib64/python2.7/io.py", line 51, in <module>
    import _io
ImportError: /usr/local/src/sage-config/local/lib64/python2.7/lib-dynload/_io.so: undefined symbol: _PyErr_ReplaceException
Error while executing Python code.
```

It seems that the system Python is used, together with the runtime files from Sage, resulting in a conflict.

vbraun:
Is that with a self-compiled gdb, or how else does gdb end up with references to /usr/local/src/sage-config? There are a number of other issues with the Sage gdb package, e.g. it just doesn't work at all on OSX....

jdemeyer:
> Is that with a self-compiled gdb

No, it's my system `gdb`.

> how else does gdb end up with references to /usr/local/src/sage-config?

I don't know really. I guess that Python somehow searches for its own runtime files and finds the one of Sage before the system ones.

vbraun:
It seems that there are only two ways this could go wrong, either gdb links to the wrong libpython:

    $ ldd $(which gdb) | grep python
	    libpython3.5m.so.1.0 => /lib64/libpython3.5m.so.1.0 (0x00007ff4b88c5000)

or sys.path is wrong inside the embedded python interpreter:

    $ gdb
    (gdb) python import sys; print(sys.path);
    ['/usr/share/gdb/python', '/usr/lib64/python35.zip', '/usr/lib64/python3.5', '/usr/lib64/python3.5/plat-linux', '/usr/lib64/python3.5/lib-dynload', '/usr/lib64/python3.5/site-packages', '/usr/lib/python3.5/site-packages']
    (gdb) python import Cython; print(Cython.__file__)
    /usr/lib64/python3.5/site-packages/Cython/__init__.py

Though both work for me; The Cython backtrace fails because Python 2<->3 incompatibility but thats to be expected.

jdemeyer:
Got it.

This problem is the `python` in `argv[0]` here:
```
        # Run gdb but pretending to be Python in argv[0].
        # This is needed for Python to find its own files.
        cmd = Popen(["python"], executable=whichgdb,
                stdin=PIPE, stdout=PIPE, stderr=PIPE, env=env)
```
Replacing `python` with `gdb` fixes the problem.

vbraun:
Ah good catch, thanks!

PR #59, opened 2017-07-20
testgdb.py is not currently distributed

kiwifb:
As it says, the file testgdb.py is not currently included when you produce a tarball. Which means that `make check-gdb` cannot be run from the produced tar.gz.


jdemeyer:
I consider `testgdb.py` to be broken anyway (see #52). So, if you want to distribute it, please add a comment in that file saying that it's known to be broken in certain cases.

kiwifb:
Ha! Fair enough, I will look into the brokeness first. I had just noticed that the makefile was referring a file not in the source distribution - or generated.

PR #64, opened 2018-01-12
add Trove classifiers

johnyf:
As of `cysignals == 1.6.5`, the POSIX requirement is [invisible on PyPI](https://pypi.org/project/cysignals/1.6.5/). Based on [this discussion](https://stackoverflow.com/a/34994245/1959808), `pip` is unaware of what value the `platforms` argument has (e.g., `pip show -v cysignals`):

https://github.com/sagemath/cysignals/blob/acc2ba9fcf8ffa3db1c0f1cf0da8deae7035f41d/setup.py#L200

Moreover, the [Trove classifier](https://pypi.python.org/pypi?%3Aaction=list_classifiers) `Operating System :: POSIX` is suitable in this case, so following [PEP 345](https://www.python.org/dev/peps/pep-0345/#platform-multiple-use), the argument `platforms` can be omitted.

My choice of "development status" was nearly arbitrary; it probably needs to be adjusted.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,6 +50,21 @@ extensions = [
     Extension("cysignals.pselect", ["src/cysignals/pselect.pyx"], **kwds),
     Extension("cysignals.tests", ["src/cysignals/tests.pyx"], **kwds),
 ]
+classifiers = [
+    'Development Status :: 3 - Alpha',
+    'Intended Audience :: Developers',
+    ('License :: OSI Approved :: '
+     'GNU Lesser General Public License v3 or later (LGPLv3+)'),
+    'Operating System :: POSIX',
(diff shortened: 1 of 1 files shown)


jdemeyer:
I'm happy to add trove classifiers, but I don't agree with the content. Especially `3 - Alpha`, what makes you think that this is "alpha" quality software?

jdemeyer:
See #65 instead

johnyf:
I did not intend to imply anything by the choice of development status specifier. From (recent) reading and usage, I think that the `cysignals` package is of good quality. The choice was mainly because I don't know what future additions/changes may be planned. Thanks for updating the PR.

PR #65, opened 2018-01-12
add Trove classifiers (updated)

jdemeyer:
@johnyf Do you agree with this?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,6 +52,26 @@ extensions = [
 ]
 
 
+classifiers = [
+    'Development Status :: 5 - Production/Stable',
+    'Intended Audience :: Developers',
+    ('License :: OSI Approved :: '
+     'GNU Lesser General Public License v3 or later (LGPLv3+)'),
+    'Operating System :: POSIX',
+    'Programming Language :: C',
+    'Programming Language :: Cython',
+    'Programming Language :: Python',
+    'Programming Language :: Python :: 2',
+    'Programming Language :: Python :: 2.7',
+    'Programming Language :: Python :: 3',
+    'Programming Language :: Python :: 3.4',
+    'Programming Language :: Python :: 3.5',
+    'Programming Language :: Python :: 3.6',
+    'Topic :: System',
+    'Topic :: Software Development :: Debuggers',
+]
+
+
 def write_if_changed(filename, text):
     """
     Write ``text`` to ``filename`` but only if it differs from the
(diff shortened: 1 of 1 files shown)


johnyf:
Yes, thanks for updating the commit. It may be worth including Python versions, though if specific versions are irrelevant, then `'Programming Language :: Python'` would communicate the same information.

jdemeyer:
> It may be worth including Python versions

I wonder why you did that actually. I don't really see the point of explicitly listing various Python versions, especially since cysignals uses very little Python functionality... it also implies that I will need to update that list when Python 3.7 or 3.8 comes out.

johnyf:
The choice of Python versions in #64 was based on `.travis.yml`. In the past I used to not include Python versions. Due to errors caused by dependencies (not here---in other packages), I started using Python versions, including both major and minor version numbers.

An example where the error was convoluted to uncover is (the commit message there contains more details):

https://github.com/tulip-control/tulip-control/commit/16d82b4fe18e144e0d342174956a38d3be32fcae

which led to removing the specifier for Python 3.3 from the classifiers of `matplotlib`, because Python 3.4 or higher was required by `matplotlib`, from the Python 3 series (https://github.com/matplotlib/matplotlib/issues/9085).

Since then, I have been first adding specific Python versions to the test environments on Travis CI (e.g., [this commit](https://github.com/tulip-control/tulip-control/commit/88e0574223319499b27061f3f688482a5aa79cb8)), and after the tests pass, declare support via Trove classifiers (e.g., [this commit](https://github.com/tulip-control/tulip-control/commit/b91fa3ff91d9dc4fd84981ff4d356c5a49f2eb91)).

So I have been using Trove classifiers for Python versions as the specification of the package, so that the information be visible on PyPI. Even if contiguous Python versions are supported, explicitly indicating the supported Python versions emphasizes that they are supported (when absent one could be unclear whether the Python version specifiers have been omitted or not considered at all).

As commented [above](https://github.com/sagemath/cysignals/pull/65#issuecomment-357182558), the `cysignals` package is mostly Python independent. I did not know whether that is the case, so I used the information from `.travis.yml`. (Also, `cysignals` requires Cython, and Cython lists specific Python versions [in its `setup.py`](https://github.com/cython/cython/blob/634a86f26c2da8fe37ee27d49c4d5be410376f59/setup.py#L264-L271), so those specifiers may be relevant here.)

Indeed, the need for updating the Python version classifiers increases logistics, so if the versions make little difference, then omitting them may simplify things. On the other hand, new Python versions will be added to `.travis.yml` as they become supported, so those commits could also add them to the classifiers, thus declaring support to `pip` and other packages.

jdemeyer:
OK, I agree. At least it communicates that it needs Python 2.7+ or 3.4+

jdemeyer:
If you agree, then I'll merge this. Note that I amended your commit, so you will show as author in git. That is why I am asking you for confirmation.

johnyf:
Thanks. I agree with the changes in the amended commit, and with merging into branch `master`.

PR #78, opened 2018-03-20
[WIP] Run CI using Conda

jdemeyer:


jdemeyer:
CC @saraedum

jdemeyer:
What's up with conda-forge and Python 3.4? Has support for that version been dropped?

saraedum:
Looks very reasonable. Can travis run tests on Linux & OSX?

jdemeyer:
> Can travis run tests on Linux & OSX?

Yes, Linux and OS X, both x86_64.

jdemeyer:
@isuruf: I am trying again to get this working. Unfortuntaly, PARI doesn't work: do I need to do anything special to link against `-lpari` within Conda?

isuruf:
@jdemeyer, you'll have to set `LDFLAGS="-L$PREFIX/lib -Wl,-rpath,$PREFIX/lib` and `CPPFLAGS=-I$PREFIX/include`

jdemeyer:
OK, I was assuming that Conda would set something like that automatically.

isuruf:
Should have been $CONDA_PREFIX instead of $PREFIX.

jdemeyer:
Thanks! This seems to be working now.

jdemeyer:
Thanks @isuruf !

PR #79, opened 2018-03-22
New function sig_occurred() to check pending exception

jdemeyer:
CC @embray 

This is meant to implement the cysignals part of https://trac.sagemath.org/ticket/24986

### DIFF
--- a/src/cysignals/signals.pyx
+++ b/src/cysignals/signals.pyx
@@ -5,7 +5,7 @@ See ``tests.pyx`` for extensive tests.
 """
 
 #*****************************************************************************
-#       Copyright (C) 2011-2016 Jeroen Demeyer <J.Demeyer@UGent.be>
+#       Copyright (C) 2011-2018 Jeroen Demeyer <J.Demeyer@UGent.be>
 #
 #  cysignals is free software: you can redistribute it and/or modify it
 #  under the terms of the GNU Lesser General Public License as published
@@ -26,8 +26,15 @@ from __future__ import absolute_import
 
 from libc.signal cimport *
 from libc.stdio cimport freopen, stdin
-from cpython.exc cimport PyErr_Occurred, PyErr_SetString
+from cpython.ref cimport Py_XINCREF, Py_XDECREF
+from cpython.exc cimport (PyErr_Occurred, PyErr_NormalizeException,
+        PyErr_Fetch, PyErr_Restore)
+from cpython.version cimport PY_MAJOR_VERSION
+
 cimport cython
+import sys
+from gc import collect
+
 
 cdef extern from "implementation.c":
     cysigs_t cysigs
@@ -38,6 +45,12 @@ cdef extern from "implementation.c":
     void _sig_on_interrupt_received() nogil
     void _sig_on_recover() nogil
     void _sig_off_warning(const char*, int) nogil
+
+    # Python library functions for raising exceptions without "except"
+    # clause.
+    void PyErr_SetNone(object type)
+    void PyErr_SetString(object type, char *message)
+    void PyErr_Format(object exception, char *format, ...)
 
 
 class AlarmInterrupt(KeyboardInterrupt):
@@ -62,6 +75,7 @@ class AlarmInterrupt(KeyboardInterrupt):
 
     """
     pass
+
 
 class SignalError(BaseException):
     """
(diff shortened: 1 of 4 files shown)


src/cysignals/signals.pyx
```diff
@@ -261,3 +284,45 @@ def python_check_interrupt(sig, frame):
...
+        except AttributeError:
+            pass
+        else:
+            if <PyObject*>handled is cysigs.exc_value:
+                Py_XDECREF(cysigs.exc_value)
+                cysigs.exc_value = NULL
+                return
+
+    # To be safe, we run the garbage collector because it may clear
+    # references to our exception.
```
embray: I'd have to think about an actual example, but couldn't an exception wind up in a reference cycle, but also be in `sys.last_value`,  in which the above case should be considered again too?  Perhaps this could install call recursively back into `verify_exc_value` with a flag set to not run garbage collection a second time.

src/cysignals/signals.pxd
```diff
@@ -51,6 +57,28 @@ cdef inline void cython_check_exception() nogil except *:
...
+    are done handling the exception, return ``NULL``.
+
+    This is meant for Cython code to check whether objects may be in
+    an invalid state. Typically, this would be used in an ``except``
+    or ``finally`` block or in ``__dealloc__``.
+
+    The implementation is based on reference counting to check whether
+    the exception has been deleted. This means that it will break if the
+    exception is stored somewhere or if the exception is involved in a
+    reference cycle.
```
embray: Naming: "interrupt_occurred" is a bit confusing; from the machine level yes it's an interrupt, but in the context of signal handling this naming could be confusing with SIGINT.  It also doesn't have the `sig_` prefix of other Cysignals functions.  So I was thinking just something like `sig_occurred()` indicating that Cysignals is handling an exception (somewhat analogous to `PyErr_Occurred`).

Should it really be considered "broken" in that case?  Really it just means a change in semantics--this doesn't *necessarily* mean that the returned exception is currently being handled (though it probably is in most cases).  Instead it's just returning the last exception raised by Cysignals that still has references to it (whether because it's being handled, or was stored in a variable for some reason).

To this end it might also be useful to have a `sig_clear()` to manually do `Py_XDECREF(cysigs.exc_value); cysigs.exc_value = NULL`.  Then, even if we did want to store the exception in a variable, for example, a wary user could do something like:

```
err = sig_occurred()
sig_clear()
... do stuff with err ...
```

Does that make any sense or am I off base?

src/cysignals/signals.pyx
jdemeyer: > I'd have to think about an actual example, but couldn't an exception wind up in a reference cycle, but also be in sys.last_value, in which the above case should be considered again too?

If the exception is `sys.last_value`, the refcount doesn't matter: we assume that the exception has been handled.

jdemeyer: Speaking of garbage collection, I'm not sure whether we should actually do that. If we are calling `sig_occurred` a lot of times with an active exception, that will be a massive performance loss.

It all depends how likely it is that an exception will end up in a reference cycle.

src/cysignals/signals.pxd
jdemeyer: > Should it really be considered "broken" in that case?

Well, the ideal semantic of `sig_occurred` is: return the exception from cysignals if it has not been handled yet. Of course, now it comes down to defining what "handled" means, but intuitively it means that the `try/except/finally` block containing `sig_on` has been exited.

jdemeyer: A more likely scenario to add a reference to the exception would be
```
typ, val, tb = sys.exc_info()
... do stuff with typ, val, tb ...
```
i.e. the code doing something with the exception is most likely unaware of cysignals.

src/cysignals/signals.pyx
embray: I see what you're saying about `sys.last_value`.  At that point we don't care anymore.

embray: On the issue of garbage collection--it's a shame the gc module doesn't expose an API to check for a given object if it's reachable from outside the GC, without actually performing a full collection.

If nothing else the import statement should be moved outside this function--there's no reason for it to be inside.  We could also avoid an unnecessary full collection by checking `gc.get_referrers(cysigs.exc_value)`.  That doesn't necessarily guarantee that it's part of a reference cycle or that `collect()` would do anything to affect its reference count (e.g. it could be part of a cycle that includes an object with `__del__`).   But if it has no referrers tracked by the GC then obviously it's unreachable.

embray: Such a check wouldn't alleviate the case you mentioned though where the exception is being handled and there are other objects referring to it.  Then repeated calls to `sig_occurred` would still be slow.

jdemeyer: > But if it has no referrers tracked by the GC then obviously it's unreachable.

I disagree here. Not every object reference is tracked by GC. For example, Cython local variables wouldn't be and I doubt that `sys.exc_info` is.

So, I don't see how `get_referrers` would solve anything.

jdemeyer:
Branch rebased with some minor updates...

Any further comments?

PR #82, opened 2018-03-24
Update .travis.yml

isuruf:


jdemeyer:
This still fails on Python 3.4 because Cython 0.28 is not packaged for Python 3.4 and it's not clear to me why.

PR #83, opened 2018-03-26
Better handling on Cygwin for exceptions that occur during exception handling on an alt stack

### LINKED ISSUE #77: Test failures on Cygwin
Running `make test` on Cygwin I get these failures:

```
**********************************************************************
File "src/cysignals/tests.pyx", line 533, in tests.pyx
Failed example:
    print(Popen([executable, '-c', cmd], stdout=PIPE, stderr=PIPE).communicate()[1].decode("utf-8"))
Expected:
    ------------------------------------------------------------------------
    ...
    ------------------------------------------------------------------------
    <BLANKLINE>
Got:
    ------------------------------------------------------------------------
    <BLANKLINE>
**********************************************************************
File "src/cysignals/tests.pyx", line 581, in tests.pyx
Failed example:
    print(msg.decode("utf-8"))
Expected:
    ------------------------------------------------------------------------
    ...
    ------------------------------------------------------------------------
    Unhandled SIG...
    This probably occurred because a *compiled* module has a bug
    in it and is not properly wrapped with sig_on(), sig_off().
    Python will now terminate.
    ------------------------------------------------------------------------
    <BLANKLINE>
Got:
    ------------------------------------------------------------------------
    Unhandled SIGSEGV: A segmentation fault occurred.
    This probably occurred because a *compiled* module has a bug
    in it and is not properly wrapped with sig_on(), sig_off().
    Python will now terminate.
    ------------------------------------------------------------------------
    <BLANKLINE>
**********************************************************************
File "src/cysignals/tests.pyx", line 632, in tests.pyx
Failed example:
    print(Popen([executable, '-c', cmd], stdout=PIPE, stderr=PIPE).communicate()[1].decode("utf-8"))
Expected:
    ------------------------------------------------------------------------
    ...
    ------------------------------------------------------------------------
    Unhandled SIGABRT: An abort() occurred.
    This probably occurred because a *compiled* module has a bug
    in it and is not properly wrapped with sig_on(), sig_off().
    Python will now terminate.
    ------------------------------------------------------------------------
    <BLANKLINE>
Got:
    ------------------------------------------------------------------------
    Unhandled SIGABRT: An abort() occurred.
    This probably occurred because a *compiled* module has a bug
    in it and is not properly wrapped with sig_on(), sig_off().
    Python will now terminate.
    ------------------------------------------------------------------------
    <BLANKLINE>
**********************************************************************
File "src/cysignals/tests.pyx", line 672, in tests.pyx
Failed example:
    print(Popen([executable, '-c', cmd], stdout=PIPE, stderr=PIPE).communicate()[1].decode("utf-8"))
Expected:
    ------------------------------------------------------------------------
    ...
    ------------------------------------------------------------------------
    Unhandled SIGSEGV: A segmentation fault occurred.
    This probably occurred because a *compiled* module has a bug
    in it and is not properly wrapped with sig_on(), sig_off().
    Python will now terminate.
    ------------------------------------------------------------------------
    <BLANKLINE>
Got:
    ------------------------------------------------------------------------
    Unhandled SIGSEGV: A segmentation fault occurred.
    This probably occurred because a *compiled* module has a bug
    in it and is not properly wrapped with sig_on(), sig_off().
    Python will now terminate.
    ------------------------------------------------------------------------
    <BLANKLINE>
**********************************************************************
File "src/cysignals/tests.pyx", line 697, in tests.pyx
Failed example:
    print(Popen([executable, '-c', cmd], stdout=PIPE, stderr=PIPE).communicate()[1].decode("utf-8"))
Expected:
    ------------------------------------------------------------------------
    ...
    ------------------------------------------------------------------------
    An error occurred during signal handling.
    This probably occurred because a *compiled* module has a bug
    in it and is not properly wrapped with sig_on(), sig_off().
    Python will now terminate.
    ------------------------------------------------------------------------
    <BLANKLINE>
Got:
    <BLANKLINE>
**********************************************************************
1 items had failures:
   5 of 120 in tests.pyx
***Test Failed*** 5 failures.
```

Seems pretty trivial though.  The tests are expecting some extra `-------` followed by some text that isn't being output.  I'm not sure why.

embray:
This is a rough idea for how to fix #77.  Actually it seems to work pretty reliably but I admit I haven't fully worked out the implications for continuing to do anything with Cygwin in this environment.  It's probably pretty unsafe, but this is still better than the current situation (where the OS just quietly terminates the process, and never even sets the exit code in a way that Cygwin can keep track of).

### DIFF
--- a/src/cysignals/implementation.c
+++ b/src/cysignals/implementation.c
@@ -133,6 +133,7 @@ static inline void sig_reset_defaults(void) {
  * whether the exception occurred inside our signal handler */
 static inline void sigdie_for_sig(int sig, int inside)
 {
+    sig_reset_defaults();
     if (inside) sigdie(sig, "An error occurred during signal handling.");
 
     /* Quit Python with an appropriate message. */
@@ -249,7 +250,6 @@ static void cysigs_signal_handler(int sig)
 
         /* Reset all signals to their default behaviour and unblock
          * them in case something goes wrong as of now. */
-        sig_reset_defaults();
         sigdie_for_sig(sig, inside);
     }
 }
--- a/src/cysignals/implementation.c
+++ b/src/cysignals/implementation.c
@@ -546,8 +546,11 @@ static void sigdie(int sig, const char* s)
      * the user is probably using other debugging tools and we don't
      * want to interfere with that. */
 #else
-#ifndef __APPLE__
+#if !(defined(__APPLE__) || defined(__CYGWIN__))
     /* See http://trac.sagemath.org/13889 for how Apple screwed this up */
+    /* On Cygwin this has never quite worked, and in particular when run
+       from the altstack handler it just results in fork errors, so disable
+       this feature for now */
     if (getenv("CYSIGNALS_CRASH_NDEBUG") == NULL)
         print_enhanced_backtrace();
 #endif
--- a/src/cysignals/implementation.c
+++ b/src/cysignals/implementation.c
@@ -114,6 +114,54 @@ static inline void reset_CPU(void)
 }
 
 
+/* Reset all signal handlers and the signal mask to their defaults. */
+static inline void sig_reset_defaults(void) {
+    signal(SIGHUP, SIG_DFL);
+    signal(SIGINT, SIG_DFL);
+    signal(SIGQUIT, SIG_DFL);
+    signal(SIGILL, SIG_DFL);
+    signal(SIGABRT, SIG_DFL);
+    signal(SIGFPE, SIG_DFL);
+    signal(SIGBUS, SIG_DFL);
+    signal(SIGSEGV, SIG_DFL);
+    signal(SIGALRM, SIG_DFL);
+    signal(SIGTERM, SIG_DFL);
+    sigprocmask(SIG_SETMASK, &default_sigmask, NULL);
+}
+
+
+/* Call sigdie() with the appropriate error message for the signal, or
+ * whether the exception occurred inside our signal handler */
+static inline void sigdie_for_sig(int sig, int inside)
+{
+    if (inside) sigdie(sig, "An error occurred during signal handling.");
+
+    /* Quit Python with an appropriate message. */
+    switch(sig)
+    {
+        case SIGQUIT:
+            sigdie(sig, NULL);
+            break;  /* This will not be reached */
+        case SIGILL:
+            sigdie(sig, "Unhandled SIGILL: An illegal instruction occurred.");
+            break;  /* This will not be reached */
+        case SIGABRT:
+            sigdie(sig, "Unhandled SIGABRT: An abort() occurred.");
+            break;  /* This will not be reached */
+        case SIGFPE:
+            sigdie(sig, "Unhandled SIGFPE: An unhandled floating point exception occurred.");
+            break;  /* This will not be reached */
+        case SIGBUS:
+            sigdie(sig, "Unhandled SIGBUS: A bus error occurred.");
+            break;  /* This will not be reached */
+        case SIGSEGV:
+            sigdie(sig, "Unhandled SIGSEGV: A segmentation fault occurred.");
+            break;  /* This will not be reached */
+    };
+    sigdie(sig, "Unknown signal received.\n");
+}
+
+
 /* Handler for SIGHUP, SIGINT, SIGALRM
  *
  * Inside sig_on() (i.e. when cysigs.sig_on_count is positive), this
@@ -196,43 +244,8 @@ static void cysigs_signal_handler(int sig)
 
         /* Reset all signals to their default behaviour and unblock
          * them in case something goes wrong as of now. */
-        signal(SIGHUP, SIG_DFL);
-        signal(SIGINT, SIG_DFL);
-        signal(SIGQUIT, SIG_DFL);
-        signal(SIGILL, SIG_DFL);
-        signal(SIGABRT, SIG_DFL);
-        signal(SIGFPE, SIG_DFL);
-        signal(SIGBUS, SIG_DFL);
-        signal(SIGSEGV, SIG_DFL);
-        signal(SIGALRM, SIG_DFL);
-        signal(SIGTERM, SIG_DFL);
-        sigprocmask(SIG_SETMASK, &default_sigmask, NULL);
-
-        if (inside) sigdie(sig, "An error occurred during signal handling.");
-
-        /* Quit Python with an appropriate message. */
-        switch(sig)
-        {
-            case SIGQUIT:
-                sigdie(sig, NULL);
-                break;  /* This will not be reached */
-            case SIGILL:
-                sigdie(sig, "Unhandled SIGILL: An illegal instruction occurred.");
-                break;  /* This will not be reached */
-            case SIGABRT:
-                sigdie(sig, "Unhandled SIGABRT: An abort() occurred.");
-                break;  /* This will not be reached */
-            case SIGFPE:
-                sigdie(sig, "Unhandled SIGFPE: An unhandled floating point exception occurred.");
-                break;  /* This will not be reached */
-            case SIGBUS:
-                sigdie(sig, "Unhandled SIGBUS: A bus error occurred.");
-                break;  /* This will not be reached */
-            case SIGSEGV:
-                sigdie(sig, "Unhandled SIGSEGV: A segmentation fault occurred.");
-                break;  /* This will not be reached */
-        };
-        sigdie(sig, "Unknown signal received.\n");
+        sig_reset_defaults();
+        sigdie_for_sig(sig, inside);
     }
 }
 
--- a/src/cysignals/implementation.c
+++ b/src/cysignals/implementation.c
@@ -162,6 +162,12 @@ static inline void sigdie_for_sig(int sig, int inside)
 }
 
 
+/* Additional platform-specific implementation code */
+#if defined(__CYGWIN__) && defined(__x86_64__)
+#include "implementation_cygwin.h"
+#endif
+
+
 /* Handler for SIGHUP, SIGINT, SIGALRM
  *
  * Inside sig_on() (i.e. when cysigs.sig_on_count is positive), this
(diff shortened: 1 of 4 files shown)


src/cysignals/implementation.c
```diff
@@ -107,6 +108,60 @@ static inline void reset_CPU(void)
...
+            sigdie(sig, "Unhandled SIGSEGV: A segmentation fault occurred.");
+            break;  /* This will not be reached */
+    };
+    sigdie(sig, "Unknown signal received.\n");
+}
+
+
+/* Additional platform-specific implementation code */
+#if defined(__CYGWIN__) && defined(__x86_64__)
+#include "implementation_cygwin.h"
```
jdemeyer: I find an `#include` in the middle of the code bad style. Can you put it near the top of the file?

src/cysignals/implementation.c
```diff
@@ -107,6 +108,60 @@ static inline void reset_CPU(void)
...
+        case SIGSEGV:
+            sigdie(sig, "Unhandled SIGSEGV: A segmentation fault occurred.");
+            break;  /* This will not be reached */
+    };
+    sigdie(sig, "Unknown signal received.\n");
+}
+
+
+/* Additional platform-specific implementation code */
+#if defined(__CYGWIN__) && defined(__x86_64__)
```
jdemeyer: Why `defined(__x86_64__)`? Even if you need that, it makes more sense to put that check in `implementation_cygwin.h`

src/cysignals/implementation_cygwin.h
```diff
@@ -0,0 +1,61 @@
+/* Cygwin-specific implementation details */
```
jdemeyer: This should be a `.c` file instead of `.h`

src/cysignals/implementation.c
```diff
@@ -107,6 +108,60 @@ static inline void reset_CPU(void)
 }
 
 
+/* Reset all signal handlers and the signal mask to their defaults. */
+static inline void sig_reset_defaults(void) {
```
jdemeyer: Given that calls to `sig_reset_defaults` and `sigdie_for_sig` always appear together, why make them separate functions? They are logically related because I need to reset the signal handlers for the `kill` in `sigdie` to work.

src/cysignals/implementation.c
```diff
@@ -344,6 +364,7 @@ static void setup_alt_stack(void)
     ss.ss_size = sizeof(alt_stack);
     ss.ss_flags = 0;
     if (sigaltstack(&ss, NULL) == -1) {perror("sigaltstack"); exit(1);}
+    PLATFORM_SETUP_ALT_STACK;
```
jdemeyer: I don't like `PLATFORM_SETUP_ALT_STACK`.

I would rather write this as
```
#ifdef __CYGWIN__
    cygwin_setup_alt_stack();
#endif
```
just to be explicit that something is happening here only on Cygwin.

jdemeyer:
It would be interesting to know what happens when you don't call `setup_alt_stack` at all. For all we know, that might just work.

src/cysignals/implementation_cygwin.h
embray: That was my original call as well--it was a ".c" file to start with and I had `#include "implementation_cygwin.c"`.  However, I found out the hard way that in this case re-running `setup.py` didn't rebuild `implementation.o`, which is annoying.  Open to ideas.

src/cysignals/implementation.c
embray: Yes, I agree--I'll change that.

src/cysignals/implementation.c
embray: I agree, though though the file's contents depends on some inline functions I added above.  Without it I get complaints about missing declarations--in that case it might almost be better to add an `implementation.h` and move the declarations there.

src/cysignals/implementation.c
embray: Okay--I was sort of split on that because `sig_reset_defaults` looked like it could be useful on its own as well.  But in practice it only ended being used in this one place.  Alternatively I could keep `sig_reset_defaults` as a separate routine, but place a call to it at the beginning of `sigdie`.

src/cysignals/implementation.c
embray: My thinking was in theory other platforms might add something here, but I'm just as happy to do the above.

embray:
> For all we know, that might just work.

No, it won't.  Cygwin's behavior here is in line with Linux--if a stack overflow occurs it actually disables any non-default SIGSEGV handler unless we have an alternative stack.

src/cysignals/implementation_cygwin.h
jdemeyer: > I found out the hard way that in this case re-running setup.py didn't rebuild implementation.o, which is annoying.

That issue is independent of the filename.

src/cysignals/implementation.c
jdemeyer: Even if multiple platforms need to do something, I would prefer platform-specific `#ifdef`s.

src/cysignals/implementation.c
jdemeyer: > Without it I get complaints about missing declarations

Just add declarations in the beginning of `implementation.c`.

> in that case it might almost be better to add an implementation.h

My feeling is that header files are mostly used to define a public API, which is not the case here.

embray: That's not necessarily true at all--headers are just so that declarations an be shared between compilation units, and doesn't necessarily have anything to do with public APIs.  It's not unusual to have headers that are used internally for build purposes, but are never "installed".

src/cysignals/implementation_cygwin.h
embray: You would think, but when I changed the filename that seemed to fix it.  Maybe it was something else.  
I think that, rather than doing an "include" I'd rather just make this a separate (optional) compilation unit.  However, that will require adding a couple header files.

embray:
Going to see if I can get this cleaned up.  After re-reading this, it probably does make sense to still have a separate `implementation_cygwin.{c,h}` since there's enough code in it that it would be a lot of noise to include in the main `implementation.c`.

embray:
Finally got around to updating this.  I think it will be a bit better now.

We still need to add CI for Cygwin, and I can do that.  For now you'll have to I guess take my word for it that this works.

embray:
I moved the call to `sig_reset_defaults` inside `sigdie_for_sig`--I agree it doesn't make sense to not call them together.  But they still sort of make sense to exist as separate functions.

jdemeyer:
The `__version__` stuff that you added is breaking on Python 2.7 according to Travis CI.

embray:
Oops, I don't think I meant to add that stuff here at all.  I think they were just uncommitted changes that I accidentally `git ci -a`'d

embray:
Ugh. It might be unrelated to this, but now I'm getting a hang when I run `make check` that didn't happen before I rebased on current master.

embray:
Nevermind, I think might be user error.  However, now I'm getting a new failure:

```
File "src/cysignals/signals.pyx", line 181, in signals.pyx
Failed example:
    init_cysignals()
Expected:
    <cyfunction python_check_interrupt at ...>
Got:
    <built-in function python_check_interrupt>
```

embray:
Still user/environment error.  But now I'm back to the tests hanging in `src/cysignals/tests.pyx`.  I'm going to try going back to the master branch instead of mine and see if it's still a problem.

embray:
This is weird: `make check` currently runs `make check-prefix` and then `make check-user`.  If I, starting from a clean repository, run both of those targets in order, manually, everything works.  However, if I just run `make check`, again from a clean repo, it hangs in the `check-user-doctest` target on:

```
export PYTHONUSERBASE="`pwd`/tmp/user" && ulimit 2>/dev/null -s 1024; python -B rundoctests.py src/cysignals/*.pyx
Doctesting 5 files.
src/cysignals/alarm.pyx
src/cysignals/pselect.pyx
src/cysignals/pysignals.pyx
src/cysignals/signals.pyx
src/cysignals/tests.pyx
```

I think some of these makefile rules really need to be cleaned up in general, but it's disconcerting that something about how the makefile is written can actually cause the tests to hang.  So I need to keep looking into it.  This behavior seems to be deterministic though.

embray:
Alright. After more extensive testing over lunch, it looks like this might not be entirely deterministic after all, and maybe it has nothing to do with the different make targets as my initial testing suggested.  So now I just need to find out exactly what test it's hanging on...

embray:
Okay. When it does hang it seems to be happening in `test_interrupt_bomb`.   This is hard to reproduce, but I suspect that's what happening is that the "Some time later" is too short for Cygwin, considering the additional overhead of signal handling there, and the `SIGABRT` is getting lost in the shuffle.  I don't know that that *should* happen, and it might indicate a bug in Cygwin.  But I don't think there's anything to fix here other than to make the test a little more robust somehow (i.e. ensure it can't block forever).

jdemeyer:
> When it does hang it seems to be happening in test_interrupt_bomb.

Interesting that you get a hang, that indicated that the `SIGABRT` is completely ignored. That seems to me like the least likely outcome (I could understand a hard crash).

I would be interested to see the output of this test when cysignals was configured with `./configure --enable-debug`.

Does it help if you increase the delay in
```
signal_after_delay(SIGABRT, base_delay + 10*n + 1000)
```

jdemeyer:
Note that the `test_interrupt_bomb` is really about the interaction between the many `SIGINT`s. The `SIGABRT` is just meant to terminate the loop, it's not really considered part of the test. So there might be a genuine bug here that the `SIGINT` interacts with the `SIGABRT`.

jdemeyer:
Reading the cysignals code, it seems that the case of two different signals arriving together isn't really supported well. So if you get a `SIGABRT` while handling a `SIGINT`, the `SIGABRT` might be swallowed.

This might be fixable, but it's not completely obvious because there are many race conditions to think about. Even then, the `SIGABRT` would occur outside of the `sig_on` and would just crash the test.

So I should probably change the `SIGABRT` mechanism in this test to some delay.

embray:
> Reading the cysignals code, it seems that the case of two different signals arriving together isn't really supported well. So if you get a SIGABRT while handling a SIGINT, the SIGABRT might be swallowed.

Yes, that's what I was implying in [this comment](https://github.com/sagemath/cysignals/pull/83#issuecomment-386274467).  It's entirely possible for the `SIGABRT` to just never be seen by cysignals under such heavy bombardment.

But anyways this discussion might be better continued elsewhere since it's not actually pertinent to this ticket (it just came up here because I didn't start seeing this failure until I most recently tested this ticket).

embray:
Thanks!

PR #91, opened 2018-05-03
Skip test_interrupt_bomb on Cygwin

embray:


embray:
This still doesn't work, because `doctest` still runs the doctest even if the `test_interrupt_bomb` function is not defined.

embray:
I see. That's what @vinklein added here: https://github.com/sagemath/cysignals/pull/76/files#diff-f2dd0623f1b5311c41b53a1fc96707f6  

I wonder if maybe we couldn't pull that out to a separate PR so I can depend on it.

vinklein:
Sure we can. The necessity of doing it depends on when PR#76 will be merged and when do you need it.
I am just being aware that a separate PR will not have any sense by itself (adding doctest flag for windows when there is no windows version)

jdemeyer:
Keep the function but change the test instead.

jdemeyer:
I just pushed a commit to master to fix `test_interrupt_bomb`. Can you try it?

embray:
It seems that this test still hangs on Cygwin.  I'm not sure why at this point since the reworking in 9d2a437ebf4913bd3d4ef1b6829934deac158112 looks like it should have at least prevented it from looping forever.  Now I get "Doctest src/cysignals/tests.pyx terminated. Timeout limit exceeded (>600s)" which is better than before but still not what one would expect.  It very well might be a bug--or at least limitation--in Cygwin.

embray:
Rebased on #94 for now.

embray:
Oops, I thought this was merged already but apparently not.

embray:
?

jdemeyer:
Fixed in master now.

PR #94, opened 2018-05-14
Allow doctests to be skipped by platform

embray:
This feature was added in #76.  This PR refactors the change from that PR into a separate change.  I also moved stuff around a bit such that the implementation depends less on underscored internals of the doctest module.  Also added Cygwin as a platform for which tests can be skipped (originally for #91, though that might not be necessary anymore since 9d2a437ebf4913bd3d4ef1b6829934deac158112).

### DIFF
--- a/rundoctests.py
+++ b/rundoctests.py
@@ -9,7 +9,7 @@
 import os
 import sys
 import doctest
-from doctest import DocTestParser, OPTIONFLAGS_BY_NAME
+from doctest import DocTestParser, Example, SKIP
 from multiprocessing import Process
 if os.name != 'nt':
     import resource
@@ -19,38 +19,40 @@ timeout = 600
 
 filenames = list(sys.argv[1:])
 if os.name == 'nt':
-    notinwindows = ['src/cysignals/pysignals.pyx', 'src/cysignals/alarm.pyx', 'src/cysignals/pselect.pyx']
+    notinwindows = ['src/cysignals/pysignals.pyx',
+                    'src/cysignals/alarm.pyx',
+                    'src/cysignals/pselect.pyx']
 
     for f in list(filenames):
         if f in notinwindows:
             filenames.remove(f)
 
 # Add an option to flag doctest what should not be run on windows.
-doctest.register_optionflag("SKIP_WINDOWS")
-doctest.register_optionflag("SKIP_POSIX")
-flag_skip_windows = OPTIONFLAGS_BY_NAME["SKIP_WINDOWS"]
-flag_skip_posix = OPTIONFLAGS_BY_NAME["SKIP_POSIX"]
+SKIP_WINDOWS = doctest.register_optionflag("SKIP_WINDOWS")
+SKIP_CYGWIN = doctest.register_optionflag("SKIP_CYGWIN")
+SKIP_POSIX = doctest.register_optionflag("SKIP_POSIX")
+
+flag_map = {
+    SKIP_WINDOWS: lambda: os.name == 'nt',
+    SKIP_CYGWIN: lambda: sys.platform == 'cygwin',
+    SKIP_POSIX: lambda: os.name == 'posix'
+}
 
 
-class SkipByOsDocTestParser(DocTestParser):
+class CysignalsDocTestParser(DocTestParser):
+    def parse(self, *args, **kwargs):
+        examples = DocTestParser.parse(self, *args, **kwargs)
+        for example in examples:
+            if not isinstance(example, Example):
+                continue
+            for flag, flag_test in flag_map.items():
+                if flag in example.options and flag_test():
+                    example.options[SKIP] = True
 
-    def _find_options(self, source, name, lineno):
-        options = DocTestParser._find_options(self, source, name, lineno)
-
-        if flag_skip_windows in options.keys() and os.name == 'nt':
-            # Replace SKIP_WINDOWS with SKIP flag if we are on windows.
-            options[OPTIONFLAGS_BY_NAME["SKIP"]] = options[flag_skip_windows]
-            del options[flag_skip_windows]
-
-        if flag_skip_posix in options.keys() and os.name != 'nt':
-            # Replace SKIP_POSIX with SKIP flag if we are not on windows.
-            options[OPTIONFLAGS_BY_NAME["SKIP"]] = options[flag_skip_posix]
-            del options[flag_skip_posix]
-
-        return options
+        return examples
 
 
-parser = SkipByOsDocTestParser()
+parser = CysignalsDocTestParser()
 
 
 print("Doctesting {} files.".format(len(filenames)))
--- a/rundoctests.py
+++ b/rundoctests.py
@@ -9,11 +9,50 @@
 import os
 import sys
 import doctest
-import resource
+from doctest import DocTestParser, OPTIONFLAGS_BY_NAME
+from multiprocessing import Process
+if os.name != 'nt':
+    import resource
 
 flags = doctest.ELLIPSIS
+timeout = 600
 
-filenames = sys.argv[1:]
+filenames = list(sys.argv[1:])
+if os.name == 'nt':
+    notinwindows = ['src/cysignals/pysignals.pyx', 'src/cysignals/alarm.pyx', 'src/cysignals/pselect.pyx']
+
+    for f in list(filenames):
+        if f in notinwindows:
+            filenames.remove(f)
+
+# Add an option to flag doctest what should not be run on windows.
+doctest.register_optionflag("SKIP_WINDOWS")
+doctest.register_optionflag("SKIP_POSIX")
+flag_skip_windows = OPTIONFLAGS_BY_NAME["SKIP_WINDOWS"]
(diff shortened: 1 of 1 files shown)


embray:
@vinklein if this is acceptable we might want to rebase #76 on top of it.  I want to get that finally finished up.

rundoctests.py
```diff
@@ -9,11 +9,52 @@
...
+SKIP_POSIX = doctest.register_optionflag("SKIP_POSIX")
+
+flag_map = {
+    SKIP_WINDOWS: lambda: os.name == 'nt',
+    SKIP_CYGWIN: lambda: sys.platform == 'cygwin',
+    SKIP_POSIX: lambda: os.name == 'posix'
+}
+
+
+class CysignalsDocTestParser(DocTestParser):
```
embray: Note, I renamed this from its original name `SkipByOsDocTestParser` to something more general just in anticipation of possibly adding additional doctest flags particular for Cysignals.

vinklein:
I am on it.

vinklein:
```
if flag in example.options and flag_test():
    example.options[SKIP] = True
```
It's way nicer than my previous version : )

#76 has been rebased on top of #94, it works.

embray:
@jdemeyer from https://github.com/sagemath/cysignals/pull/95#issuecomment-388819562

> Psychologically, I find it easier to list the conditions for running a doctest in the positive sense: I prefer > "run this test only if X" instead of "skip this test if not X". In SageMath, it also works like that: we say # > optional - PKG for "run this test only if PKG is available`.
>
> So I would prefer flags like `OPTIONAL_POSIX` or `POSIX_ONLY`.

I'm not so sure about that.  In the Python world at least it seems more common to mark tests with some "skip on condition".  This is true of both the Python unittest module, and frameworks like nose and pytest.  You'll see this far more commonly.

That said, there are other cases where I agree with the `POSIX_ONLY` approach, especially as a doctest flag.

In the case of #91 though, for example, it makes much more sense to have an explicit `SKIP_CYGWIN` flag.  In this it's explicitly excluding a platform due to a known(?) issue on that platform.

embray:
@vinklein Thanks!  Depending on what @jdemeyer says you may have to do it again unfortunately.  But it's still useful to have your PR as a test case since it's more non-trivial.

embray:
For example, for the Windows-specific doctests I agree with @jdemeyer it would be nice to have a `WINDOWS_ONLY` flag and similarly a `POSIX_ONLY`.  I'll work on that.  Then it would just be a matter of negation...

vinklein:
Im ok with tag like `POSIX_ONLY` but against `OPTIONAL_POSIX`. The "optional" word sounds misleading to me in this case.

rundoctests.py
```diff
@@ -9,11 +9,52 @@
 import os
 import sys
 import doctest
-import resource
+from doctest import DocTestParser, Example, SKIP
+from multiprocessing import Process
+if os.name != 'nt':
+    import resource
```
jdemeyer: Move this import inside the block below where you actually use it.

rundoctests.py
```diff
@@ -9,11 +9,52 @@
...
 flags = doctest.ELLIPSIS
+timeout = 600
+
+filenames = list(sys.argv[1:])
+if os.name == 'nt':
+    notinwindows = ['src/cysignals/pysignals.pyx',
+                    'src/cysignals/alarm.pyx',
+                    'src/cysignals/pselect.pyx']
+
+    for f in list(filenames):
```
jdemeyer: No need for explicit `list()`

rundoctests.py
```diff
@@ -24,47 +65,55 @@
...
+
+        status = p.exitcode
+
+        if p.is_alive():
+            p.terminate()
+            print("Doctest {} terminated. Timeout limit exceeded (>{}s)".format(f, timeout))
+            success = False
+        elif status != 0:
+            success = False
+            if os.name != 'nt':
```
jdemeyer: We still want to print something if `os.name == 'nt'` too.

rundoctests.py
```diff
@@ -24,47 +65,55 @@
...
+        if sys.platform == 'darwin':
+            from cysignals.signals import _setup_alt_stack
+            _setup_alt_stack()
+        failures, _ = doctest.testfile(file, module_relative=False, optionflags=flags, parser=parser)
+        if not failures:
+            os._exit(0)
+    finally:
+        os._exit(23)
+
+if __name__ == "__main__": # Mandatory for windows cases.
```
jdemeyer: `windows cases` sounds strange to me. If you do put a comment, I would prefer `# Needed for Windows` or something like that. Alternatively: this construction is standard enough, I don't think that it needs a comment.

rundoctests.py
```diff
@@ -24,47 +65,55 @@
...
+        print(f)
+        sys.stdout.flush()
+
+        # Test every file in a separate process (like in SageMath) to avoid
+        # side effects from doctests.
+        p = Process(target=testfile, args=(f,))
+        p.start()
+        p.join(timeout)
+
+        status = p.exitcode
```
jdemeyer: This doesn't work the same way as the return value from `os.waitpid`.

jdemeyer: It seems that signals are represented by a negative value and that normal exit codes are just passed as-is. So the logic below needs to be fixed.

vinklein: This patch seems to work for me.
```
diff --git a/rundoctests.py b/rundoctests.py
index 4d6bd73..3c5179d 100755
--- a/rundoctests.py
+++ b/rundoctests.py
@@ -106,14 +106,9 @@ if __name__ == "__main__": # Mandatory for windows cases.
         elif status != 0:
             success = False
             if os.name != 'nt':
-                if os.WIFEXITED(status):
-                    st = os.WEXITSTATUS(status)
-                    if st != 23:
-                        print("bad exit: {}".format(st))
-                elif os.WIFSIGNALED(status):
-                    sig = os.WTERMSIG(status)
-                    print("killed by signal: {}".format(sig))
-                else:
-                    print("unknown status: {}".format(status))
+                if status < 0:
+                    print("killed by signal: {}".format(abs(status)))
+                elif status != 23:
+                    print("bad exit: {}".format(status))
```

rundoctests.py
embray: I don't like that, but if you insist.

embray: I take it back--in this case it's only used at module-level anyways so I agree this should be merged into that block.

rundoctests.py
embray: Maybe you meant the previous `list()`.  Here it's necessary since `filenames` is modified in the loop.

embray:
Also I was not fine-grained in interpreting the return codes from tests on Windows.  Perhaps @vinklein would have a better idea about how it should be interpreted, if at all.

rundoctests.py
jdemeyer: Yes, something like that should work.

vinklein:
The `if os.name != 'nt':` is here for `os.WIF*` in the first place. The current code might work on windows without that , i will test it. Either way i think you can remove it as #94 isn't supposed to support windows by itself. I will do the necessary in #76 if needed

jdemeyer:
Shouldn't the `else` branch with `print("unknown status: {}".format(status), file=sys.stderr)` be removed?

Can you remove the `os.name != 'nt'` condition for checking the exit status? We probably don't need (and if we do, we can still change it later when we truly support Windows).

vinklein:
> Shouldn't the else branch with print("unknown status: {}".format(status), file=sys.stderr) be removed?

I agree, that's what i have proposed in my patch.

embray:
I'm not really following what the intent is here (and actually in retrospect these changes to the test runner for supporting Windows are a bit orthogonal to the skip flags).  @vinklein since you seem to know what's needed here can you make a pull request to my pull request with the proposed fixes (or simply post an exact patch here)?

jdemeyer:
I'll do that.

jdemeyer:
```diff
diff --git a/rundoctests.py b/rundoctests.py
index 7c4d0e6..9c93dfd 100755
--- a/rundoctests.py
+++ b/rundoctests.py
@@ -106,16 +106,10 @@ if __name__ == "__main__": # Mandatory for windows cases.
             success = False
         elif status != 0:
             success = False
-            if os.name != 'nt':
-                if status < 0:
-                    print("killed by signal: {}".format(abs(status)),
-                          file=sys.stderr)
-                elif status != 23:
-                    print("bad exit: {}".format(status), file=sys.stderr)
-                else:
-                    print("unknown status: {}".format(status), file=sys.stderr)
-            else:
-                print("doctest exited with error status: {}".format(status),
+            if status < 0:
+                print("killed by signal: {}".format(abs(status)),
                       file=sys.stderr)
+            elif status != 23:
+                print("bad exit: {}".format(status), file=sys.stderr)
 
     sys.exit(0 if success else 1)
```

embray:
Let's see how this does...

jdemeyer:
Superseded by #96 

PS: this is one thing I hate about GitHub

Issue #92, opened 2018-05-03
Add AppVeyor build for Cygwin


embray:
This should be done after #76 is merged, since it is already adding an AppVeyor build of Cysignals for Windows.

The AppVeyor images come with Cygwin already installed, and I have some scripts I've already written to run commands through Cygwin, so then we can just add Cygwin builds to the configuration matrix of the existing `.appveyor.yml` from #76 .

PR #95, opened 2018-05-14
Cygwin builds on AppVeyor

embray:
Implements #92 in a manner intended to be compatible with #76.  Depends on #94.

It would probably be best to implement #76 first and then merge this into its `.appveyor.yml`.  But we could also do it the other way around depending on how things go with #76.

See https://ci.appveyor.com/project/embray/cysignals for sample builds.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -17,6 +17,9 @@ environment:
       MAKE: make
 
 before_build:
+  - echo "Building Cysignals for %OS%"
+  - set PATH=%PYTHON%;%PYTHON%\\scripts;%TOOLSPATH%;%PATH%
+  - echo %PATH%
   - ps: >-
       if ( "$Env:OS" -ieq "Cygwin" ) {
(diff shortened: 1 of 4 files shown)


embray:
Went ahead and hooked up AppVeyor to this repository.  You can see the successful build of this PR below.

embray:
> What's `os.name` on Cygwin?
`'posix'`

jdemeyer:
Psychologically, I find it easier to list the conditions for running a doctest in the positive sense: I prefer "run this test only if X" instead of "skip this test if not X". In SageMath, it also works like that: we say `# optional - PKG` for "run this test only if PKG is available`.

So I would prefer flags like `OPTIONAL_POSIX` or `POSIX_ONLY`.

embray:
Ok but this would make more sense to discuss on #94.  Sorry, I should have clarified that this PR is based on that one.

PR #97, opened 2018-07-12
Run autoreconf instead of autoconf & autoheader

saraedum:
this is an attempt to upstream bits of Debian's patches, namely @jgmbenoit's `upstream-autotoolization-harden-generic.patch`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,8 +143,7 @@ distcheck: dist
 #####################
 
 configure: configure.ac
-	autoconf
-	autoheader
+	autoreconf -if
 	@rm -f src/config.h.in~
 
 .PHONY: all build doc install dist doc clean clean-build clean-doc \


jdemeyer:
Which problem does this solve?

saraedum:
Sorry, I thought that the `-i` was necessary sometimes. But actually it's not. I guess Debian can drop that one.

saraedum:
Why the `@rm` btw?

jdemeyer:
I guess that `autoheader` generates that file as backup...

saraedum:
I see. Sorry for the noise.

saraedum:
I just switched branches and did `make configure` and got
```
make configure
autoconf
configure.ac:17: error: possibly undefined macro: AC_DEFINE
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
```
Whereas `autoreconf -if` actually worked.

jdemeyer:
Well, that never happened to me :-) Is the problem reproducible?

What's your `autoconf --version` and `autoreconf --version`?

saraedum:
Both are 2.69. It's reproducible when I switch to the branch of #98. I can try to produce a more minimal example…but autoreconf seems to be made for exactly that purpose, so I wonder why we should not use it anyway?

jdemeyer:
I sometimes find that `autoreconf` does too many things that it shouldn't do, especially on a very simple autoconf-only (without automake or libtool) project. Anyway, I'm not really against your patch, but I think that there was a reason why I used `autoconf && autoheader` since I would normally use `autoreconf`.

jdemeyer:
> It's reproducible when I switch to the branch of #98.

In that case, I'd prefer to move the fix to #98 (which it is already anyway).

PR #38, opened 2016-09-15
Build machinery hardening

jgmbenoit:
These 3 patches slightly improve the building machinery.
Having said that, it would be a good idea to have checks that does not involve a system wide installation.

PR #98, opened 2018-07-12
Make PARI configurable

saraedum:
This is part of #38 which makes PARI configurable in the configure script.


saraedum:
I have not reviewed the m4 script in detail. I just want to bring this up for discussion as this can be helpful for packaging and also improves the configuration messages it seems.

saraedum:
I agree with most what you write here: I would also prefer feature checks to version checks, and this m4 script should be considered for upstreaming (not sure, maybe @jgmbenoit has tried to upstream already?)
Probably Debian is using this m4 script in other places where the `PARI_CPPFLAGS` are used?
Anyway, let me try to propose an alternative, just so we get a simpler `--without-pari` switch. I assume that this would be sufficient for most packagers.

jgmbenoit:
My aim was to write a generic pari.m4 that furnish traditional information used by autotools (so that it can be used elsewhere). My understanding is that the cysignal setup is hybrid as cysignal is: both navigate in two worlds, Python era an C era. I guess it is uncommon. I would prefer either a pure  autotools or a pure Python setup, but I guess it will request an unreasonable amount of work. The status quo is fine for me as Debian maintainer. 
Otherwise, a --without-pari switch will be just fine.

jgmbenoit:
I am agree, I should submit the pari.m4 to PARI/GP , I will see what I can do.

saraedum:
Ok. I pushed a new attempt to implement a `--with(out)-pari` switch. What do you think @jdemeyer ?

jdemeyer:
Please remove the ugly `x` in the tests like
```
test "x$with_pari" != "xno"
```

Apart from that, this looks good on first sight. I still need to test it.

jgmbenoit:
I am trying to package version 1.7.2 for debian. I can see that the setup machinery was modify,  and that it does  cause some trouble: I get
src/cysignals/signals.pyx:33:4: 'cysigs_t' is not a type identifier
Have you any hint from it can from ?

jdemeyer:
Can you post the complete build output?

jdemeyer:
Also make sure to use Cython version >= 0.28

jgmbenoit:
All is fine when I build the vanilla version. The issue pops up when I build the package. Is there any way to get debug messages ?

jdemeyer:
> The issue pops up when I build the package.

What does Debian do differently from the vanilla version?

jdemeyer:
If you want help, you will need to give me more details...

saraedum:
@jdemeyer I removed the "x" prefixes. Anything else that you need from my end for this?

saraedum:
@jdemeyer sorry but I could I get some feedback here maybe?

PR #102, opened 2018-12-13
Cygwin builds on AppVeyor

jdemeyer:
CC @embray 

Unfortunately failing with
```
gcc -fno-strict-aliasing -ggdb -O2 -pipe -Wimplicit-function-declaration -fdebug-prefix-map=/usr/src/ports/python2/python2-2.7.14-1.x86_64/build=/usr/src/debug/python2-2.7.14-1 -fdebug-prefix-map=/usr/src/ports/python2/python2-2.7.14-1.x86_64/src/Python-2.7.14=/usr/src/debug/python2-2.7.14-1 -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -I/cygdrive/c/projects/cysignals/tmp/site-packages/cysignals -I/usr/include/python2.7 -c cysignals_example.cpp -o build/temp.cygwin-2.10.0-x86_64-2.7/cysignals_example.o
gcc: error: spawn: No such file or directory
```

This may be relevant: https://stackoverflow.com/questions/16621529/gcc-fails-with-spawn-no-such-file-or-directory

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,4 +1,4 @@
-version: 1.0.{build}
+version: "build #{build}"
 
 build: off
 
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,48 @@
+version: 1.0.{build}
+
+build: off
+
+environment:
+  matrix:
+
+    # For Python versions available on Appveyor, see
+    # http://www.appveyor.com/docs/installed-software#python
+    # The list here is complete (excluding Python 2.6, which
+    # isn't covered by this document) at the time of writing.
+    - os: Cygwin
+      CYG_ROOT: "C:\\cygwin64"
+      PYTHON: "C:\\cygwin64\\bin"
+      PYTHON_VERSION: "2.7"
+      ARCH: x86_64
+      MAKE: make
+
+before_build:
+  - echo "Building Cysignals for %OS%"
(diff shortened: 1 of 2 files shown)


embray:
Is the C++ example file new?  Might just be missing the right g++ package...

jdemeyer:
> Is the C++ example file new?

The example is not new, but the fact that it is C++ (as opposed to C) might be new.

jdemeyer:
Minor question: what does this mean in `.appveyor.yml`:
```
version: 1.0.{build}
```
That's confusing because now I see for example "cysignals 1.0.11" in the AppVeyor page.

jdemeyer:
@embray, you should have write access to this github repo, you can just amend your commits.

embray:
It's basically boilerplate
https://www.appveyor.com/docs/build-configuration/#build-versioning

It's not the version of the software but a "build version". I think how we use it is kind of up to us, and I can just as easily do away with the `1.0` part and just use:

`version: {build}`.

jdemeyer:
Is there any particular reason that you are testing only Python 2.7 and not some Python 3.x?

jdemeyer:
I am asking because we're having some problems on #76 with silly errors from AppVeyor like
```
export PYTHONPATH="`pwd`/tmp/site-packages" && cd example && python setup.py clean build
/usr/bin/sh: python: command not found
```

PR #104, opened 2019-02-12
Cysignals for windows

jdemeyer:
Modify cysignals for windows compatibility

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -9,7 +9,23 @@ environment:
     # http://www.appveyor.com/docs/installed-software#python
     # The list here is complete (excluding Python 2.6, which
     # isn't covered by this document) at the time of writing.
-    - os: Cygwin
+    - OS: Windows_NT
+      PYTHON: "C:\\Python27-x64"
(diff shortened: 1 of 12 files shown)


embray:
I saw an old comment you left mentioning trouble testing in Python 3.  Is that still being a problem?  For Cygwin I want to add testing for at least 3.6 and maybe 3.7 now that there's a 3.7 package for Cygwin.  I don't think there was any specific reason I left it off for now, except maybe just to reduce the problem space.

jdemeyer:
> I saw an old comment you left mentioning trouble testing in Python 3. Is that still being a problem?

No, as far as I know, there are no problems with Python 3. The remaining problems are all related to the fact that signal handling on Windows is a very thing from POSIX :-)

There are still some tests to be fixed before this is ready to merge.

jdemeyer:
I just realized that "a different thread from the main thread receives a signal" is something that I'll need to fix anyway for #109. So I'll probably generalize the `sig_mapped_to_FPE` thing.

tobiasdiez:
What's the status of this PR? Windows support would be very welcome and would probably fix https://github.com/sagemath/cysignals/issues/119 as well.

kliem:
Maybe it would also be a good step for windows, to get rid of `configure.ac`?

PR #106, opened 2019-02-27
get-pip.py now writes a warning if run with python2

### LINKED ISSUE #105: get-pip.py seems to fail on Cygwin
See [AppVeyor log](https://ci.appveyor.com/project/sagemath/cysignals/builds/22315257).

CC @embray

embray:
cygwin has python2-pip and python3-pip packages now so maybe it's best
to just use the system package

attempting to fix #105

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -22,7 +22,7 @@ before_build:
   - echo %PATH%
   - ps: >-
       if ( "$Env:OS" -ieq "Cygwin" ) {
-          $python = "python" + $Env:PYTHON_VERSION[0]
+          $python = "python" + $Env:PYTHON_VERSION[0] + $Env:PYTHON_VERSION[2]
           Start-Process -NoNewWindow -Wait `
               -FilePath $Env:CYG_ROOT\setup-$Env:ARCH.exe `
               -ArgumentList "-q -P $python,$python-devel,$python-pip,gcc-core,gcc-g++"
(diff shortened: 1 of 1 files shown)


embray:
Unsurprisingly, this will likely take a few iterations.  Now it just fails with

{{{
python%PYTHON_VERSION% -m pip install -r requirements.txt
/usr/bin/python2.7: No module named pip
}}}

I wonder if the Cygwin on here has the pip package yet...

jdemeyer:
Isn't there something analogous to `get-pip.py 2>/dev/null`?

embray:
I believe so, but it should still be possible just to install the cygwin package for pip.  I'm not sure why it isn't working.

embray:
@jdemeyer This seems to be working now.  I can squash the commits if you'd prefer.

jdemeyer:
Is this going to break once Python 3.10 is released?
```
$python = "python" + $Env:PYTHON_VERSION[0] + $Env:PYTHON_VERSION[2]
```

embray:
Maybe it should use a substring instead.  I'm not sure there's going to be a Python 3.10 though. I think the plan is to go 3.9 -> 4.0.

jdemeyer:
>  I'm not sure there's going to be a Python 3.10 though. I think the plan is to go 3.9 -> 4.0.

At this point, I don't think that there is a plan at all. I also heard opinions that want to avoid Python 4 because it might make people think that stuff will be broken badly, just like Python 2 -> Python 3.

embray:
That might be wise, actually.  Given the effort that's gone into just *packaging* "python3".  I would just consider Python 4.0 to be "python3" version 3.9(+0.1), but that would be confusing to most people.

PR #109, opened 2019-03-06
Better support for threads

jdemeyer:
This pull request aims to improve thread-safety of cysignals. There are many aspects to it and this does not claim to fix all thread-related issues. In particular, we assume that there is only 1 thread any time that `sig_on()` is called. In particular, we don't support Python multithreading.

### DIFF
--- a/src/cysignals/implementation.c
+++ b/src/cysignals/implementation.c
@@ -177,6 +177,105 @@ static inline void sigdie_for_sig(int sig, int inside)
 #endif
 
 
+/* This is used to deal with multi-threaded programs. We make a difference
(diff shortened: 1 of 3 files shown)


jdemeyer:
Testing this is going to be an interesting challenge: the obvious thing to do would be OpenMP, but apparently GCC doesn't support atomics in combination with OpenMP: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65467

jdemeyer:
Now struggling with pesky `$MACOSX_DEPLOYMENT_TARGET` issues...

jdemeyer:
I merged part of this branch already, but keeping it open since there are more thread issues to be fixed.

PR #127, opened 2021-03-05
Fix #126

### LINKED ISSUE #126: possible bug in verify_exc_value
This is to track an issue I found while working on [this ticket](https://trac.sagemath.org/ticket/31404#comment:20).  So far I have only been able to reproduce the bug on the branch for that ticket, though I don't think it's directly related to anything else in that ticket (not outside the realm of possibility).  Nevertheless I think there is a bug.

From what I can tell so far at a high level this is what's happening.  The test where the bug is occurring looks like:

```
sage: _ = singular._expect.sendline('1+')  # unfinished input ## line 503 ##
sage: try:
    alarm(0.5)
    singular._expect_expr('>')  # interrupt this
except KeyboardInterrupt:
    pass ## line 504 ##
Control-C pressed. Interrupting Singular. Please wait a few seconds...
```

though it could in principle be any code using `alarm()`.  Then what happens is:

1. The alarm triggers a `SIGINTR` which gets handled by cysignals.
2. We are not in a `sig_on()` block, so `PyErr_SetInterrupt()` is called to call the Python interrupt handler.
3. This correctly sets the `AlarmInterrupt` exception and also sets `cysigs.exc_value` to the exception value.
4. Here I'm not sure, but it seems the signal was received during garbage collection or something, because when the Python interpreter resumes it tries deallocating an Integer, and hits the `sig_occurred()` check which is performed before putting a possibly broken `mpz_t` on the free list.  So this happens before Python has raised the `AlarmInterrupt` exception.
5. I found that the first time this happens, `cysigs.exc_value.ob_refcnt == 3`.
6. This results in running the garbage collector again [here](https://github.com/sagemath/cysignals/blob/2886287e2c20fa1e04bd8ace3449949ea01ce86b/src/cysignals/signals.pyx#L368).
7. This results in more integers deallocated which again goes into `sig_occurred()`.  This happens a few more times until...
8. At some point `cysigs.exc_value.ob_refcnt` is reduced to 1 *after* the call to `collect()` (I don't know why the exception is being derefed and that needs to be looked into more).  This results in `cysigs.exc_value` set to `NULL`.
9. We return to the other `collect()` call (remember, we are in nested garbage collect calls; I have no idea if this is even safe to do...)
10. After the outer `collect()` call, `verify_exc_value` again checks the value of `cysigs.exc_value.ob_refcnt` and segfaults since `cysigs.exc_value` is `NULL`.

In summary there are at least two issues here:

1. Why is the exception being derefed?
2. `verify_exc_value` should make sure `cysigs.exc_value` is not set to `NULL` after garbage collection, but before checking its ref count.
3. (Question) Is it even OK to call `gc.collect()` recursively?

embray:
Includes a regression test.  Without this fix, the test causes the
doctest runner itself to segfault.  Not sure if that's ideal, but it
does at least demonstrate the bug, and that it's fixed by this change.

### DIFF
--- a/src/cysignals/signals.pyx
+++ b/src/cysignals/signals.pyx
@@ -372,6 +372,9 @@ cdef void verify_exc_value():
         # is not functional anymore.
         pass
 
-    if cysigs.exc_value.ob_refcnt == 1:
+    # Make sure we still have cysigs.exc_value at all; if this function was
+    # called again during garbage collection it might have already been set
+    # to NULL; see https://github.com/sagemath/cysignals/issues/126
+    if cysigs.exc_value != NULL and cysigs.exc_value.ob_refcnt == 1:
         Py_XDECREF(cysigs.exc_value)
         cysigs.exc_value = NULL
--- a/src/cysignals/tests.pyx
+++ b/src/cysignals/tests.pyx
@@ -1016,6 +1016,7 @@ def test_sig_occurred_dealloc_in_gc():
     keep a reference to the exception so it doesn't go away right away::
 
         >>> from cysignals.tests import *
+        >>> import sys
         >>> e = None
         >>> try:
         ...     test_sig_occurred_dealloc_in_gc()
@@ -1024,7 +1025,12 @@ def test_sig_occurred_dealloc_in_gc():
         >>> print_sig_occurred()
         RuntimeError: test_sig_occurred_dealloc_in_gc()
 
-    Put the exception into a dict containing a reference to itself, so that
+    Python 2 keeps the target of the "except as" in local scope, so make sure
+    to delete it as well::
+
+        >>> if sys.version_info[0] < 3: del exc
+
+    Put the exception into a list containing a reference to itself, so that
     when the garbage collector runs (in ``verify_exc_value``) its reference
     count drops to 1.  Also include a ``DeallocDebug`` so that
     ``sig_occurred()`` is called during GC.
@@ -1033,11 +1039,11 @@ def test_sig_occurred_dealloc_in_gc():
     collector is not called except by ``verify_exc_value()``::
 
         >>> import gc
-        >>> d = {'e': e, 'x': DeallocDebug()}
-        >>> d['d'] = d
+        >>> l = [DeallocDebug(), e]
+        >>> l.append(l)
         >>> gc.disable()
         >>> try:
-        ...     del d, e
+        ...     del l, e
         ...     print_sig_occurred()
         ... finally:
         ...     gc.enable()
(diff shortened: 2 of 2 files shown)


embray:
Test is failing on Python 2.7, possibly because its design depends in part on dict ordering (and the order in which dict elements are deallocated after breaking a reference cycle).  That's just my guess.  On older Python 3 versions they are failing for some other  reason I think unrelated to this PR.

malb:
This looks good to me. Shall we just remove the Python 2 tests, it's EOL?

embray:
@malb See #132 and also #133 where I'm trying to fix the CI for older Python versions at least for now.

embray:
Re-running tests now that #133 is merged.

dimpase:
Wow! Can we have a release, for testing purposes?

embray:
@dimpase Yeah, I'll make a pre-release soon and put a ticket on Sage to test it.

kliem:
There seems to be some problem with this. I'm testing this on https://trac.sagemath.org/ticket/31455.
I'm getting a segementation fault with

`sage -t --long --random-seed=0 src/sage/matrix/matrix_mod2_dense.pyx`

with the test `sage: MatrixSpace(GF(2), 2^30)(1)  # optional - memlimit ## line 171 ##`

I attached traceback
[cysignals_matrix_mod2.log](https://github.com/sagemath/cysignals/files/6111097/cysignals_matrix_mod2.log)

I don't think this is related to anything I did, because the file `sage/local/lib/python3.7/site-packages/cysignals/cysignals_config.h` is the only thing that I efficiently changed and in my case it is equivalent as `-fopenmp` does not make a difference.

collares:
@kliem Quoting #131 comment 0:

<blockquote>
[1.10.3 will not contain] the assembly longjmp implementation, which should at a minimum necessitate a new minor release since it's a big addition (and currently does not work for me in Sage, but that's another story).
</blockquote>

The longjmp changes were committed back in 2019, but won't be included in 1.10.3. Could they be causing your issue?

kliem:
I don't know what's causing the issue. I figured it was this, because there are only two closed pull requests, but I didn't see that before then there commits without pull requests.

collares:
@embray created a tag for 1.10.3b0 at https://github.com/sagemath/cysignals/releases/tag/1.10.3b0. Do you see the same problem using this version?

embray:
@kliem Try testing cysignals with the 1.10.x branch which does not include the longjmp changes (which I've observed still cause a lot of problems in Sage)

slel:
The "releases" page https://github.com/sagemath/cysignals/releases
does list 1.10.3, but somehow still shows 1.10.2 as the "Latest release".

dimpase:
indeed, that's cause the release title was not filled it. Fixed now.

PR #128, opened 2021-03-05
Compilation with fopenmp

### LINKED ISSUE #122: struct_signals.h cannot be compiled with -fopenmp
From https://trac.sagemath.org/ticket/31245#comment:11

The following from struct_signals.h 
does not work with `-fopenmp`:

    #if CYSIGNALS_C_ATOMIC || CYSIGNALS_CXX_ATOMIC
     typedef volatile _Atomic int cy_atomic_int;

It fails compiling:

    [sagelib-9.3.beta6] /var/lib/buildbot/slave/sage_git/build/local/lib/python3.9/site-packages/cysignals/struct_signals.h:45:1: sorry, unimplemented: ‘_Atomic’ with OpenMP
    [sagelib-9.3.beta6]  typedef volatile _Atomic int cy_atomic_int;

This and probably the next check should probably appended by checking whether `_OPENMP` is compiled.

kliem:
This is supposed to fix #122. I'm not entirely sure how to test it though.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -130,6 +130,22 @@ AC_COMPILE_IFELSE([AC_LANG_SOURCE(
     dnl YES
     [AC_MSG_RESULT([yes])]
     AC_DEFINE(CYSIGNALS_C_ATOMIC, 1, [Define to 1 if your C compiler supports _Atomic.])
+
+    AC_MSG_CHECKING([for _Atomic with OpenMP in C code])
+    saved_CFLAGS="$CFLAGS"
+    CFLAGS="$CFLAGS -fopenmp"
+    AC_COMPILE_IFELSE([AC_LANG_SOURCE(
+        [
+        static _Atomic int x;
+        ])],
+        dnl YES
+        [AC_MSG_RESULT([yes])]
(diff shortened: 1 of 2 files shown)


embray:
How do you reproduce the problem in the first place?  Do you just add `CFLAGS=-fopenmp` when compiling?

kliem:
> How do you reproduce the problem in the first place? Do you just add `CFLAGS=-fopenmp` when compiling?

I don't know exactly. It happens with sage's kucalc builtbot:

https://trac.sagemath.org/ticket/31245#comment:11

Probably I should try reproducing this with tox/github actions.

embray:
I see, so this doesn't really impact cysignals itself, but rather using cysignals macros in third-party code that is being compiled with OpenMP support.  It also seems to be compiler-specific as the relevant issue appears to be [fixed in GCC](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65467) (though I admit I don't fully understand what's being fixed here).

One concern about this is that by having configure-time checks it will disable use of _Atomic for all code using cysignals, with or without OpenMP.  I'm not sure there's a good workaround to that...

embray:
Or is that what the `#if _OPENMP` is supposed to address?

Issue #132, opened 2021-03-08
Drop unsupported Python versions


embray:
It might be good to drop support for older Python versions (at least <3.6), and maybe also add support for newer versions (3.8 and 3.9)

avishmehta68710:
is this issue closed now

embray:
No, not at all.  I've been laying some groundwork, but nothing has been done to finalize it.

embray:
Done in #142.

PR #133, opened 2021-03-08
Fix CI on older Python versions

embray:
There are some CI issues with Python 2.7 and Python 3.4.  As noted in #132 we should drop support for these, but in the meantime (at least for 1.10.x releases) I would like to get them working again.


embray:
Great, it's working now.

PR #136, opened 2021-03-09
Run `twine check` on the source tarball during CI checks.

embray:
Add missing long_description_content_type keyword in setup.py.


embray:
Rebased this on top of #137 which is needed for any hope of the Python 3.4 jobs passing.

PR #137, opened 2021-03-09
Automatically confirm any prompts from conda install

embray:
This should fix failing builds like https://travis-ci.com/github/sagemath/cysignals/jobs/489407955

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,10 @@ install:
     - conda install -y setuptools pip cython sphinx pari
     # The 'pip' package for Python 3.4 on conda is broken for newer versions
     - if [ "$PYTHON_VERSION" = "3.4" ]; then conda install -y 'pip<20.0'; fi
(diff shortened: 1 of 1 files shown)


embray:
This seems to be better now, but there is still a problem with the Python 3.4 builds, in that conda installs Cython 0.27, but cysignals has its Cython dependency set to 0.28, so even with `pip install --no-deps` it still builds a newer Cython to satisfy `setup_requires`, making the builds take extra long.  I wonder if it would be OK to downgrade to Cython 0.27 at a minimum.

embray:
In fact the Cython>=0.28 dependency is required for the build to work, and would be a pain to undo just for supporting Python 3.4.

Cython 0.28 does have wheels for Python 3.4, so maybe I can just have it install those.

embray:
Using pip to upgrade to Cython 0.29 worked on Linux, but not on MacOS.  Apparently this is a bug that did exist at one time but is fixed in newer pip versions: https://github.com/pypa/pip/issues/3801

PR #138, opened 2021-03-09
Check for macros correctly

kliem:
Many macros are either `1` or undefined.

Instead of

    #if CYSIGNALS_ASM_CYSETJMP

we should correctly check

    #ifdef CYSIGNALS_ASM_CYSETJMP

See also: https://github.com/sagemath/cysignals/pull/134#discussion_r590435141

### DIFF
--- a/src/cysignals/cysetjmp.h
+++ b/src/cysignals/cysetjmp.h
@@ -26,7 +26,7 @@
 #include <setjmp.h>
 
 
-#if CYSIGNALS_ASM_CYSETJMP
+#ifdef CYSIGNALS_ASM_CYSETJMP
 #ifdef __x86_64__
 /*
  * x86_64 assembly implementation of cysetjmp(): we store the registers
--- a/src/cysignals/implementation.c
+++ b/src/cysignals/implementation.c
@@ -23,7 +23,7 @@ Interrupt and signal handling for Cython
  ****************************************************************************/
 
 
-#if __USE_FORTIFY_LEVEL
+#ifdef __USE_FORTIFY_LEVEL
 #error "cysignals must be compiled without _FORTIFY_SOURCE"
 #endif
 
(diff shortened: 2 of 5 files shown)


collares:
From the C11 spec:

<blockquote>
Preprocessing directives of the forms

``` 
#if constant-expression new-line group opt
#elif constant-expression new-line group opt
``` 
check whether the controlling constant expression evaluates to nonzero.

Prior to evaluation, macro invocations in the list of preprocessing tokens that will become the controlling constant expression are replaced (except for those macro names modified by the defined unary operator), just as in normal text. If the token defined is generated as a result of this replacement process or use of the defined unary operator does not match one of the two specified forms prior to macro replacement, the behavior is undefined. After all replacements due to macro expansion and the defined unary operator have been performed, all remaining identifiers (including those lexically identical to keywords) are replaced with the pp-number 0, and then each preprocessing token is converted into a token."
</blockquote>

So using #if turns undefined identifiers into 0 by spec.

embray:
@collares Is that true for older specs, like C99?  I was wondering about this as well, since I am usually (if not due to cargo cult) careful to specify `#ifdef`.

kliem:
Tell me, if I should close this. I don't have any strong feelings either way.

collares:
In my opinion (which doesn't matter, but I was the one being pedantic so I think I should clarify what I meant), clarity is better than exploiting the spec. So if you think `#ifdef` is clearer, I think this is a worthwhile change even if it isn't necessary.

embray:
It depends on the macro.  The `HAVE_*_H` macros are either defined or not.  Likewise any other macros filled in by autoconf are marked `#undef` by default in the template file, so they are always either undefined or defined to 1.  I think `#if HAVE_...` reads a bit more intuitively though.

PR #142, opened 2021-03-17
Drop Python 2 support

embray:
There are really very few Python 2-isms in the main cysignals code.
Mostly just some __future__ imports and one or two other bits.

Almost to the point where it's not worth dropping completely;
nevertheless better not to encourage Python 2 use.  The main point of
this is also to drop support from the CI.

For now I kept language_level=2 for the Cython modules, but we could
probably change it to 3 as well.  There are very few places in cysignals
where this matters much.

Finally, added a few f-strings in the pure Python code where
appropriate, now that it's possible.


embray:
The one travis failure looks like a fluke, but Travis is so broken these days, trying to restart a job fails every time I try.

Since the build for the PR is exactly the same thing, and it passed, that's good enough for me.

PR #144, opened 2021-03-17
Try to fix AppVeyor builds

### LINKED ISSUE #143: Fix AppVeyor build for Cygwin
The AppVeyor build has still been running for a while, but not reporting back to GitHub since the change to GitHub "Apps".

I have re-enabled it, but it appears to be failing when trying to run Python in Cygwin with an unhelpful:

```
if ( "$Env:OS" -ieq "Cygwin" ) {
    $python = "python" + $Env:PYTHON_VERSION[0] + $Env:PYTHON_VERSION[2]
    Start-Process -NoNewWindow -Wait `
        -FilePath $Env:CYG_ROOT\setup-$Env:ARCH.exe `
        -ArgumentList "-q -P $python,$python-devel,$python-pip,gcc-core,gcc-g++,libcrypt-devel"
}
cd C:\projects\cysignals
python%PYTHON_VERSION% -c "import sys; print(sys.version); print(sys.platform); print(sys.path)"
Access is denied.
```

It used to work but that was a long time ago to be sure...

embray:
Fixes #143 

I'm not sure if this is the problem, but I think AppVeyor no longer supports "os:" (which was never well documented anyways).

Try OS as environment variable explicitly instead.

Also change the PYTHON environment variable to PYTHON_PATH which makes
its purpose a bit clearer.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -11,21 +11,24 @@ environment:
     # isn't covered by this document) at the time of writing.
     - os: Cygwin
       CYG_ROOT: "C:\\cygwin64"
-      PYTHON: "C:\\cygwin64\\bin"
+      PYTHON_PATH: "C:\\cygwin64\\bin"
       PYTHON_VERSION: "2.7"
       ARCH: x86_64
       MAKE: make
 
 before_build:
   - echo "Building Cysignals for %OS%"
-  - set PATH=%PYTHON%;%PYTHON%\\scripts;%TOOLSPATH%;%PATH%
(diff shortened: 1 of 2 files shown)


embray:
This looks to be working now.  cysetjmp is disabled on Cygwin via the configure check.

Interestingly, the configure check for cysetjmp does work on my local Windows machine, and the tests pass with it compiled.  I wonder what the difference is.  It's using the same GCC version in both cases.  Some older GCCs had problems on Windows related to SIMD registers, but the GCC being used on AppVeyor shouldn't be affected by that.

embray:
I think the problem with appveyor only affects the master branch, but I'll see next time I push to the 1.10.x branch.

Issue #148, opened 2021-07-07
undefined symbol 'PARI_SIGINT_block' on OpenBSD


dimpase:
On OpenBSD 6.9, cysignals builds, but cannot be imported:
```
$ ./sage --python
Python 3.8.8 (default, Apr 19 2021, 10:23:47) 
[Clang 10.0.1 ] on openbsd6
Type "help", "copyright", "credits" or "license" for more information.
>>> import cysignals
python3:/home/dima/sagetrac-mirror/local/lib/python3.8/site-packages/cysignals/signals.cpython-38.so: undefined symbol 'PARI_SIGINT_block'
python3:/home/dima/sagetrac-mirror/local/lib/python3.8/site-packages/cysignals/signals.cpython-38.so: undefined symbol 'PARI_SIGINT_pending'
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/dima/sagetrac-mirror/local/lib/python3.8/site-packages/cysignals/__init__.py", line 1, in <module>
    from .signals import AlarmInterrupt, SignalError, init_cysignals  # noqa
ImportError: Cannot load specified object
```
indeed, the names of these symbols are a bit weird:
```
 nm -D /home/dima/sagetrac-mirror/local/lib/libpari-tls.so.7 | grep PARI
000ed1d0 R __emutls_t.PARI_stack_limit
00c42f70 D __emutls_v.PARI_SIGINT_block
00c42f90 D __emutls_v.PARI_SIGINT_pending
00c42fd0 D __emutls_v.PARI_stack_limit
```
as opposed to Linux:
```
$ nm -D /usr/lib64/libpari-gmp-tls.so.7 | grep PARI
0000000000000744 B PARI_SIGINT_block
0000000000000740 B PARI_SIGINT_pending
0000000000000730 B PARI_stack_limit
```

So the hacks (`/* Fake PARI variables */`) in `src/cysignals/implementation.c` dealing with `PARI_SIGINT_` vars apparently fail.

Any idea how this can be fixed?

```

dimpase:
oops, sorry for noise, it's a problem with Pari headers, nothing serious.

Issue #125, opened 2021-02-21
Replace compile-time dependency on pari by run-time dispatch


mkoeppe:
The current design is not compatible with modern Python packaging, in which PEP517/518 dictate that there is no such thing as an "optional build dependency".
Also, if installing a wheel one can never be sure whether it has been built with the PARI connection or not.

This should be replaced by a hook into which PARI (via cypari/cypari2) can install itself.

embray:
I was thinking about this just yesterday.  I admit I don't 100% understand how the PARI integration is supposed to work or what it's for.  But AFAICT it could be replaced with a more generic hook.  Such a hook might be useful for other interfaces as well, such as for GAP.

embray:
If nothing else, for easier pip-install, `--without-pari` should be the *default*, rather than `--with-pari`.

mkoeppe:
Let's look into creating the hook sometime during in the Sage 9.4 development cycle.

dimpase:
I'm able to build wheels (for manylinux and macOS) for a cython module dependent on an external C++ library in https://github.com/dimpase/primecountpy
using https://github.com/pypa/cibuildwheel

OK, here we'd need to build libpari - should not make a difference.

Should I try?

mkoeppe:
No, the best way forward is to merge #149

dimpase:
Doesn't #149 need newer Cython?

dimpase:
If I recall correctly, one reason Pari gets in here is that its threads implementation is a bit funny, and if it's not called early things go bad, one gets some weird TLS-related segfaults. (But maybe it's my PTSD speaks up :-))

dimpase:
Anyhow, we just got a wheel builder in here, Pari or no Pari (the wheels we build do have Pari linked in).

kliem:
> If I recall correctly, one reason Pari gets in here is that its threads implementation is a bit funny, and if it's not called early things go bad, one gets some weird TLS-related segfaults. (But maybe it's my PTSD speaks up :-))

#149 removes exactly this dependency. Pari headers provide `PARI_SIGINT_block` and `PARI_SIGINT_pending`. This is all we need. What we aim for is somehow getting this stuff on runtime, if available.

I don't know if this is even possible and less so, if it is possible to integrate such a mechanism in `cysignals`. I do have a workaround by just passing pointers on runtime. #149 can get those pointers, provided that cypari2 has https://github.com/sagemath/cypari2/pull/109 merged.

[...]

PR #149, opened 2021-08-31
discover cypari2 on runtime instead of on configure

kliem:
This is intended to solve #148.

Usage, one must include PARI header now before cimporting from `cysignals` (if at all).
If the PARI headers are included after cimporting `cysignals`, it fails intentionally (this could be removed, but that PARI's signal blocking is ignored.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,10 +6,11 @@ os:
 
 env:
   matrix:
-    - PYTHON_VERSION=3.6
-    - PYTHON_VERSION=3.7
-    - PYTHON_VERSION=3.8
-    - PYTHON_VERSION=3.9
+    - PYTHON_VERSION=3.6 PARI=yes
+    - PYTHON_VERSION=3.7 PARI=yes
+    - PYTHON_VERSION=3.7 PARI=no
+    - PYTHON_VERSION=3.8 PARI=yes
+    - PYTHON_VERSION=3.9 PARI=yes
 
 install:
     # Install Miniconda.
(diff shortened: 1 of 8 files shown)


kliem:
One needs to modify `cypari/types.pxd` to link to pari, as now every file that includes PARI headers needs to link it as well.

kliem:
What do you mean by inconsistent opinion?

There is one problem I'm aware of: If you use a package that uses PARI, but doesn't include the headers, you wouldn't catch on that you are not paying attention to PARI's signal blocking status. However, in this case it is wrong usage (and should be mentioned somewhere, still don't know if anyone will read this). I don't know how to avoid this at all.

One thing that we could do is to rely on `cypari2`. In this case whenever `cysignals` is initialized it tries to import some function from `cypari2` that passes pointers to `PARI_SIGINT_block` and `PARI_SIGINT_pending` (typecasted, because it is a python function). Maybe that is better. (This approach would only require a few changes to the current branch).

mkoeppe:
> If you use a package that uses PARI, but doesn't include the headers, you wouldn't catch on that you are not paying attention to PARI's signal blocking status. 

Yes, exactly.

> However, in this case it is wrong usage (and should be mentioned somewhere, still don't know if anyone will read this).

Basically you are moving the burden of depending on / conditionally depending on PARI from `cysignals` to all of `cysignals`'s consumers. 
That's a non-solution.

> One thing that we could do is to rely on `cypari2`. In this case whenever `cysignals` is initialized it tries to import some function from `cypari2` that passes pointers to `PARI_SIGINT_block` and `PARI_SIGINT_pending` (typecasted, because it is a python function). Maybe that is better. 

Yes, something like this is what I meant by "runtime dispatch" in #125.

kliem:
Depending on `cypari2` is also a responsibility for the user. However, I agree that this is much nicer to communicate. Probably something like the following:

- In order for `cysignals` to respect PARI's blocking state, `cypari2 >= 2.1.3` must be available on runtime.

This is transparent and manageable also for unexperienced users.

Thinking about this, I like this much better than my other approach (even if it involves some typecasting from `volatile int*` to python int and back to `volatile int*`).

I'll do the changes tomorrow.

mkoeppe:
This is looking good to me; but shouldn't the CI be testing that it works without error if cypari2 is present in the runtime environment?

kliem:
Yes, but that won't work until cypari2 has a new release.
Of course, we could just install PARI and cypari2 for the CI and it will automatically change later.

mkoeppe:
Yes, here on this PR you could just set up the CI to use a pinned version `cypari2 @ URL`

mkoeppe:
Oh, the CI was still using Travis CI - this does not seem to run any more.

PR #151, opened 2021-09-28
Fix build when SIGSTKSZ is not a constant.

### LINKED ISSUE #150: MINSIGSTKSZ is no longer a constant
https://lists.gnu.org/archive/html/info-gnu/2021-08/msg00001.html

`MINSIGSTKSZ` is no longer a constant in some situations.

Apparently cysignals will not compile with GCC 11 then:
https://groups.google.com/g/sage-devel/c/Wz5Rhjnn4sU/m/rUd1PQbwAwAJ

kliem:
This is intended to fix #150 as done in https://github.com/rr-debugger/rr/issues/2916.

### DIFF
--- a/src/cysignals/implementation.c
+++ b/src/cysignals/implementation.c
@@ -458,6 +458,8 @@ static void setup_alt_stack(void)
     ss.ss_sp = mmap(NULL, stack_size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
     ss.ss_size = stack_size;
 #else
+    static char alt_stack[MINSIGSTKSZ + 5120 + BACKTRACELEN * sizeof(void*)];
+
     ss.ss_sp = alt_stack;
     ss.ss_size = sizeof(alt_stack);
 #endif
--- a/src/cysignals/implementation.c
+++ b/src/cysignals/implementation.c
@@ -44,6 +44,9 @@ Interrupt and signal handling for Cython
 #if HAVE_SYS_WAIT_H
 #include <sys/wait.h>
 #endif
+#if HAVE_SYS_MMAN_H
+#include <sys/mman.h>
+#endif
 #if HAVE_UNISTD_H
 #include <unistd.h>
 #endif
@@ -438,18 +441,26 @@ static void _sig_off_warning(const char* file, int line)
     print_backtrace();
 }
 
-
 static void setup_alt_stack(void)
 {
 #if HAVE_SIGALTSTACK
-    /* Static space for the alternate signal stack. The size should be
+    /* Space for the alternate signal stack. The size should be
      * of the form MINSIGSTKSZ + constant. The constant is chosen rather
      * ad hoc but sufficiently large. */
-    static char alt_stack[MINSIGSTKSZ + 5120 + BACKTRACELEN * sizeof(void*)];
-
     stack_t ss;
+#if HAVE_SYS_MMAN_H
+    /* MISIGSTKSZ is no longer a constant starting with glibc 2.34.
+     * If <sys/mmap.h> is present, we can use mmap as a workaround as
+     * done in
+     * https://github.com/rr-debugger/rr/commit/a744ceff00ffc04c9b3e9cc9d424d434c5a808b7
+     * */
+    size_t stack_size = MINSIGSTKSZ + 5120 + BACKTRACELEN * sizeof(void*);
+    ss.ss_sp = mmap(NULL, stack_size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+    ss.ss_size = stack_size;
+#else
     ss.ss_sp = alt_stack;
     ss.ss_size = sizeof(alt_stack);
+#endif
     ss.ss_flags = 0;
     if (sigaltstack(&ss, NULL) == -1) {perror("cysignals sigaltstack"); exit(1);}
 #endif
--- a/src/cysignals/implementation.c
+++ b/src/cysignals/implementation.c
@@ -449,7 +449,7 @@ static void setup_alt_stack(void)
      * ad hoc but sufficiently large. */
     stack_t ss;
 #if HAVE_SYS_MMAN_H
-    /* MISIGSTKSZ is no longer a constant starting with glibc 2.34.
+    /* MINSIGSTKSZ is no longer a constant starting with glibc 2.34.
      * If <sys/mmap.h> is present, we can use mmap as a workaround as
      * done in
      * https://github.com/rr-debugger/rr/commit/a744ceff00ffc04c9b3e9cc9d424d434c5a808b7
--- a/src/cysignals/implementation.c
+++ b/src/cysignals/implementation.c
@@ -44,9 +44,6 @@ Interrupt and signal handling for Cython
 #if HAVE_SYS_WAIT_H
 #include <sys/wait.h>
 #endif
-#if HAVE_SYS_MMAN_H
-#include <sys/mman.h>
-#endif
 #if HAVE_UNISTD_H
 #include <unistd.h>
 #endif
@@ -441,28 +438,25 @@ static void _sig_off_warning(const char* file, int line)
     print_backtrace();
 }
 
+
 static void setup_alt_stack(void)
 {
 #if HAVE_SIGALTSTACK
     /* Space for the alternate signal stack. The size should be
      * of the form MINSIGSTKSZ + constant. The constant is chosen rather
-     * ad hoc but sufficiently large. */
+     * ad hoc but sufficiently large.
+     * MINSIGSTKSZ is no longer a constant starting with glibc 2.34. */
     stack_t ss;
-#if HAVE_SYS_MMAN_H
-    /* MINSIGSTKSZ is no longer a constant starting with glibc 2.34.
-     * If <sys/mmap.h> is present, we can use mmap as a workaround as
-     * done in
-     * https://github.com/rr-debugger/rr/commit/a744ceff00ffc04c9b3e9cc9d424d434c5a808b7
-     * */
-    size_t stack_size = MINSIGSTKSZ + 5120 + BACKTRACELEN * sizeof(void*);
-    ss.ss_sp = mmap(NULL, stack_size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
-    ss.ss_size = stack_size;
-#else
+#if MINSIGSTKSZ_IS_CONSTANT
     static char alt_stack[MINSIGSTKSZ + 5120 + BACKTRACELEN * sizeof(void*)];
-
     ss.ss_sp = alt_stack;
     ss.ss_size = sizeof(alt_stack);
+#else
+    size_t stack_size = MINSIGSTKSZ + 5120 + BACKTRACELEN * sizeof(void*);
+    ss.ss_sp = malloc(stack_size);
+    ss.ss_size = stack_size;
 #endif
+    if (ss.ss_sp == NULL) {perror("cysignals malloc alt signal stack"); exit(1);}
     ss.ss_flags = 0;
     if (sigaltstack(&ss, NULL) == -1) {perror("cysignals sigaltstack"); exit(1);}
 #endif
--- a/README.rst
+++ b/README.rst
@@ -46,13 +46,7 @@ Changelog
 1.11.0 (unreleased)
 ^^^^^^^^^^^^^^^^^^^
 
-* Nothing changed yet.
-
-
-1.10.4 (unreleased)
-^^^^^^^^^^^^^^^^^^^
-
-* Nothing changed yet.
+* Fixed compilation with glib 3.34.
 
 
 1.10.3 (2021-03-16)
--- a/configure.ac
+++ b/configure.ac
@@ -27,7 +27,7 @@ CXX="$CC"
 CXXFLAGS="$CFLAGS"
 AC_PROG_CXX()
 
-AC_CHECK_HEADERS([execinfo.h sys/mman.h sys/prctl.h sys/time.h sys/wait.h windows.h sys/mman.h])
+AC_CHECK_HEADERS([execinfo.h sys/mman.h sys/prctl.h sys/time.h sys/wait.h windows.h])
 AC_CHECK_FUNCS([fork kill sigprocmask sigaltstack backtrace])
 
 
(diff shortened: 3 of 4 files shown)


src/cysignals/implementation.c
```diff
@@ -438,18 +441,26 @@ static void _sig_off_warning(const char* file, int line)
...
 #if HAVE_SIGALTSTACK
-    /* Static space for the alternate signal stack. The size should be
+    /* Space for the alternate signal stack. The size should be
      * of the form MINSIGSTKSZ + constant. The constant is chosen rather
      * ad hoc but sufficiently large. */
-    static char alt_stack[MINSIGSTKSZ + 5120 + BACKTRACELEN * sizeof(void*)];
-
     stack_t ss;
+#if HAVE_SYS_MMAN_H
+    /* MISIGSTKSZ is no longer a constant starting with glibc 2.34.
```
embray, line 452: typo: MISIGSTKSZ -> MINSIGSTKSZ

src/cysignals/implementation.c
```diff
@@ -438,18 +441,26 @@ static void _sig_off_warning(const char* file, int line)
...
+     * If <sys/mmap.h> is present, we can use mmap as a workaround as
+     * done in
+     * https://github.com/rr-debugger/rr/commit/a744ceff00ffc04c9b3e9cc9d424d434c5a808b7
+     * */
+    size_t stack_size = MINSIGSTKSZ + 5120 + BACKTRACELEN * sizeof(void*);
+    ss.ss_sp = mmap(NULL, stack_size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+    ss.ss_size = stack_size;
+#else
     ss.ss_sp = alt_stack;
     ss.ss_size = sizeof(alt_stack);
```
embray, line 462: The non-mmap case is broken now since you deleted the statically allocated `alt_stack` buffer.

I would suggest adding a configure check for whether or not `MINSIGSTKSZ` is constant, *however* if this is only going to be the case for older glibc versions moving forward, there's probably not much advantage in keeping backwards compat for the static allocation, since this is just code run at setup time.

Instead, as a fallback, we could use malloc as in the example given here: https://man7.org/linux/man-pages/man2/sigaltstack.2.html

In both cases (malloc and mmap) there should be error checking in case the allocation failed.

I'm not exactly sure where the constant 5120 comes from.  It's 5kb, which is one (typical) page size on Linux plus an extra kb (I'm guessing just as a safety margin).  It might also be that OSX requires a larger buffer, which I've seen referenced somewhere.

If we're getting rid of the static allocation we can also replace this constant with something like `sysconf(_SC_PAGESIZE) * 2` which should be sufficient.

I also like [this suggestion](https://stackoverflow.com/a/31820184/982257) of, when using `mmap`, to use `mprotect()` to create a couple extra guard pages on either end of the allocation (but use `sysconf(_SC_PAGESIZE)` instead of `getpagesize()`.

Another concern we need to look into: According to this commit ee6e81eaed82f87690da21be4ef7d3bc7b298473 we used to malloc the buffer, but that had problems on some older (?) OSX versions.  Jeroen wasn't totally clear about what the issue was (and without blame--it sounds confusing), but it apparently had something to do with fork: https://trac.sagemath.org/ticket/21509#comment:21

So actually, maybe for the sake of OSX, the static allocation should be kept as a possibility too.  That or the OSX issue is fixed upstream, and we don't need to care about older versions since Apple enforces planned obsolescence.  Still, someone with a more recent Mac should check this.

embray (changes requested):

src/cysignals/implementation.c
kliem: I resolved the non-mmap case now.

I'm completely uncertain what the best solution would be. `malloc` sounds nice and easy. `_setup_alt_stack` isn't even used as far as I can see other than in cysignals doctests. I don't know why `malloc` would not work in this case.

There is a small memory leak when calling `init_cysignals` more than once. But maybe this is fine? But could also be avoided by storing the address somewhere in `implementation.c`, so that on reinitialization it is being recycled.

enriqueartal:
I used kliem:fix_non_constant_MINSIGSTKSZ to creat a cysignals package to be able to compile Sagemath in Fedora 35. The test of `src/sage/rings/polynomial/polynomial_zmod_flint.pyx` complains about a ) in line 68 of cysignals-CSI;  with the correction  the test is successful

kliem:
> I used kliem:fix_non_constant_MINSIGSTKSZ to creat a cysignals package to be able to compile Sagemath in Fedora 35. The test of `src/sage/rings/polynomial/polynomial_zmod_flint.pyx` complains about a ) in line 68 of cysignals-CSI; with the correction the test is successful

I requested the fixes to cysignals-CSI already in #149.

enriqueartal:
A test of a giac file did not pass, with a complain to cysignals but I do not know if it is related with this fix.

./sage -t --long --warn-long 131.6 --random-seed=0 src/sage/libs/giac/__init__.py
Running doctests with ID 2021-11-06-22-37-54-f7ebb02c.
Git branch: u/mkoeppe/test_ticket__python_3_10_development_releases
Using --optional=build,dochtml,fedora,pip,sage,sage.geometry.polyhedron,sage.rings.real_double,sage_spkg
Doctesting 1 file.
sage -t --long --warn-long 131.6 --random-seed=0 src/sage/libs/giac/__init__.py
**********************************************************************
File "src/sage/libs/giac/__init__.py", line 18, in sage.libs.giac
Failed example:
    B = gb_giac(I.gens()) # random
Exception raised:
      File "/usr/local/sage-310/local/var/lib/sage/venv-python3.10/lib64/python3.10/site-packages/sage/doctest/forker.py", line 694, in _run
        self.compile_and_execute(example, compiler, test.globs)
      File "/usr/local/sage-310/local/var/lib/sage/venv-python3.10/lib64/python3.10/site-packages/sage/doctest/forker.py", line 1088, in compile_and_execute
        exec(compiled, globs)
      File "<doctest sage.libs.giac[3]>", line 1, in <module>
        B = gb_giac(I.gens()) # random
      File "/usr/local/sage-310/local/var/lib/sage/venv-python3.10/lib64/python3.10/site-packages/sage/libs/giac/__init__.py", line 118, in wrapper
        return func(*args, **kwds)
      File "/usr/local/sage-310/local/var/lib/sage/venv-python3.10/lib64/python3.10/site-packages/sage/libs/giac/__init__.py", line 353, in groebner_basis
        gb_giac = F.gbasis(list(var_names), giac_order)
      File "sage/libs/giac/auto-methods.pxi", line 6442, in sage.libs.giac.giac.GiacMethods_base.gbasis (build/cythonized/sage/libs/giac/giac.cpp:56012)
        return GiacMethods['gbasis'](self,*args)
      File "sage/libs/giac/giac.pyx", line 2102, in sage.libs.giac.giac.GiacFunction.__call__ (build/cythonized/sage/libs/giac/giac.cpp:153070)
        sig_on()
    RuntimeError: Aborted

kliem:
I think the giac failure is not related. It's just wrapped in `sig_on ... sig_off`, so it jumps back to `sig_on` and raises an exception then, which is exactly what should happen. (At least from `cysignals` point of view, giac shouldn't fail of course).

kliem:
> > I used kliem:fix_non_constant_MINSIGSTKSZ to creat a cysignals package to be able to compile Sagemath in Fedora 35. The test of `src/sage/rings/polynomial/polynomial_zmod_flint.pyx` complains about a ) in line 68 of cysignals-CSI; with the correction the test is successful
> 
> I requested the fixes to cysignals-CSI already in #149.

However, it also makes sense to have the fixes here. #149 is not as urgent as this issue.

dimpase:
should we merge this?

kliem:
> should we merge this?

I think so and make a release pretty soon. I just closed #139, so that we can run some tests.

kliem:
The testsuite fails apparently. Have to wait until the artifacts are uploaded.

More importantly my test that I added to configure does not work at all. Any help is welcome.

dimpase:
What's the branch you're testing?

dimpase:
Some of the runs seem to work, no?

kliem:
Yes, but `configure.ac` isn't doing what it is supposed to do. Fedora 35 fails with exactly the problem we are trying to fix.

enriqueartal:
> Some of the runs seem to work, no?

I can confirm that it is possible to build a running copy of sage (from 9.4 with python 3.9 and from develop with python 3.10). There are more tickets involved (to be able to build in F35) and some tests fail, but I can use it in the daily work.

kliem:
If I now change

    static char alt_stack[MINSIGSTKSZ + 5120];



> > Some of the runs seem to work, no?
> 
> I can confirm that it is possible to build a running copy of sage (from 9.4 with python 3.9 and from develop with python 3.10). There are more tickets involved (to be able to build in F35) and some tests fail, but I can use it in the daily work.

I believe that the current state of the branch is not working, because `configure.ac` is not working.

You can try this with cloning this rep and just run `make configure` and `./configure`. I'm guessing that the last test outputs:

    checking whether MINSIGSTKSZ is constant... yes

Not what I was aiming for, because on Fedora 35 it is not constant. On the other hand, if I do slight modifications, it fails for me, e.g.

    diff --git a/configure.ac b/configure.ac
    index 003a5bc..552b71a 100644
    --- a/configure.ac
    +++ b/configure.ac
    @@ -229,7 +229,7 @@ AC_MSG_CHECKING([whether MINSIGSTKSZ is constant])
     AC_COMPILE_IFELSE([AC_LANG_SOURCE(
         [
         #include <signal.h>
    -    static char alt_stack[MINSIGSTKSZ];
    +    static char alt_stack[MINSIGSTKSZ + 5120];
         ])],
         dnl YES
         [AC_MSG_RESULT([yes])]

So I'm gathering that it is not testing, what it is supposed to test.

enriqueartal:
> Yes, but `configure.ac` isn't doing what it is supposed to do. Fedora 35 fails with exactly the problem we are trying to fix.

I just bundled your branch and pass it to the build of sage in F35 (I got errors if I build it directly, there are some hacks in spkg-install.in

enriqueartal:
I do not understand what happens, I have errors when building outside sage; I attach the log of the sage installation
[cysignals_inside_sage.log](https://github.com/sagemath/cysignals/files/7581418/cysignals_inside_sage.log)

kliem:
> I do not understand what happens, I have errors when building outside sage; I attach the log of the sage installation [cysignals_inside_sage.log](https://github.com/sagemath/cysignals/files/7581418/cysignals_inside_sage.log)

Does the latest push work for you. As you probably guessed, your config.log should say:
    checking whether MINSIGSTKSZ is constant... no

enriqueartal:
Have you checked the patch in the src.rpm package for Fedora 35. There are several ones, maybe this one is relevant. It builds, though using your configure.ac, it still says it is a constant. 

--- src/cysignals/implementation.c.orig	2021-05-27 14:03:35.653774857 -0600
+++ src/cysignals/implementation.c	2021-05-27 14:19:42.225312505 -0600
@@ -445,7 +445,8 @@ static void setup_alt_stack(void)
     /* Static space for the alternate signal stack. The size should be
      * of the form MINSIGSTKSZ + constant. The constant is chosen rather
      * ad hoc but sufficiently large. */
-    static char alt_stack[MINSIGSTKSZ + 5120 + BACKTRACELEN * sizeof(void*)];
+    /* MINSIGSTKSIZE is now nonconstant. */
+    static char alt_stack[4096 + 5120 + BACKTRACELEN * sizeof(void*)];
 
     stack_t ss;
     ss.ss_sp = alt_stack;

dimpase (commented):
it seems some pairs of `[ ]` are missing in the test program. Moreover iirc include statements come apart from the rest of the source, like in a different parameter?

kliem:
> it seems some pairs of `[ ]` are missing in the test program. Moreover iirc include statements come apart from the rest of the source, like in a different parameter?

Thank you. This seems to be exactly the problem:
https://www.gnu.org/software/autoconf/manual/autoconf-2.65/html_node/Generating-Sources.html

enriqueartal:
Now it builds for me in Fedora 35. I will try now inside sage. Thanks!

kliem:
As can be observed by the checks of #152, the changes here did not break anything.

But we do fix Fedora 35 and other new systems.

dimpase:
A new release then?

kliem:
Yes, I'm just trying to figure out a two bug fixes yet. The macos test suite doesn't work at all and ubuntu-trusty fails. If there are no easy solutions to them, I'll just leave it at that.

I'm running the doctests verbose to see if I can figure out the ubuntu-trusty issue:

https://github.com/kliem/cysignals/actions/runs/1498523090

The original logs were not useful at all:

    ulimit 2>/dev/null -s 1024; python3 -B rundoctests.py src/cysignals/*.pyx
    cd example && python3 setup.py clean build
    Doctesting 5 files.
    src/cysignals/alarm.pyx
    running clean
    running build
    Compiling cysignals_example.pyx because it changed.
    [1/1] Cythonizing cysignals_example.pyx
    running build_ext
    building 'cysignals_example' extension
    creating build
    creating build/temp.linux-x86_64-3.9
    gcc -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -Wno-unused -g -O2 -g -O2 -Wp,-U_FORTIFY_SOURCE -fPIC -I/sage/local/var/lib/sage/venv-python3.9.7/lib/python+++3.9/site-packages/cysignals -I/sage/local/var/lib/sage/venv-python3.9.7/include/python3.9 -c cysignals_example.cpp -o build/temp.linux-x86_64-3.9/cysignals_example.o
    creating build/lib.linux-x86_64-3.9
    g++ -shared -Wl,-rpath-link,/sage/local/lib -L/sage/local/lib -Wl,-rpath,/sage/local/lib -Wl,-rpath-link,/sage/local/lib -L/sage/local/lib -Wl,-rpath,/sage/local/lib -Wl,-rpath+++-link,/sage/local/var/lib/sage/venv-python3.9.7/lib -Wl,-rpath,/sage/local/var/lib/sage/venv-python3.9.7/lib -L. -L/sage/local/lib -Wl,-rpath-link,/sage/local/lib -L/sage/local/+++lib -Wl,-rpath,/sage/local/lib -Wl,-rpath-link,/sage/local/lib -L/sage/local/lib -Wl,-rpath,/sage/local/lib -Wl,-rpath-link,/sage/local/lib -L/sage/local/lib -Wl,-rpath,/sage/lo+++cal/lib -Wl,-rpath-link,/sage/local/lib -L/sage/local/lib -Wl,-rpath,/sage/local/lib -g -O2 -Wp,-U_FORTIFY_SOURCE build/temp.linux-x86_64-3.9/cysignals_example.o -L/sage/local/v+++ar/lib/sage/venv-python3.9.7/lib -o build/lib.linux-x86_64-3.9/cysignals_example.cpython-39-x86_64-linux-gnu.so -lpari
    src/cysignals/pselect.pyx
    src/cysignals/pysignals.pyx
    src/cysignals/signals.pyx
    src/cysignals/tests.pyx
    make[3]: *** [check-doctest] Error 1
    make[3]: Target `check-install' not remade because of errors.

#154 attempts to fix the mac test suite problem. If it doesn't work, it doesn't work.

dimpase:
Does this happen locally on macOS too, or it's GitHub actions-only problem?

kliem:
I don't know. I don't have a macOS.

PR #152, opened 2021-09-28
First failing github tests before running the full sage install

kliem:
As travis is not working currently, it makes sense to have some checks working.


embray:
It appears we ran out of free OSS credits.  They are supposed to be replenished today, but I don't know when...

I have a fear that GitHub is going to pull the same switch as well with Actions--it's only a matter of time, once they have enough lock-in...  But in the meantime it can't hurt to have.

kliem:
Yes, #139 is another pull requests that adds github actions. It don't care very much. Whatever works at the moment should be good.
I agree that github will probably charge at some point for the Actions or put a limit on in.

kliem:
@mkoeppe 

I used this pull request to modify your test workflow. What do you think?

Currently, the tests take very long, even if there is an obvious failure.

https://github.com/sagemath/cysignals/actions/runs/1491849522

Currently, on mac the test suite fails. So what is the point on trying many different builds.

So we first try a simple build of cysignals on cygwin, ubuntu and mac-os. If the test suites pass, we start the dependent jobs and test whether it works within sage.

mkoeppe:
Yes, this is a good idea, LGTM.

Issue #71, opened 2018-03-08
make check fails on OSX with Python 3


saraedum:
With the latest 1.6.8 release, I am getting the following error on OSX with Python 3
```
Traceback (most recent call last):
  File "rundoctests.py", line 30, in <module>
    resource.setrlimit(resource.RLIMIT_STACK, (stacksize, stacksize))
ValueError: current limit exceeds maximum limit
```
the full log can be found here: https://travis-ci.org/conda-forge/cysignals-feedstock/jobs/350808690#L585

jdemeyer:
To be clear: do you mean **only** with Python 3? Or is it just that you happen to use Python 3?

saraedum:
It works with Python 2.7. But fails with 3.5 and 3.6.

saraedum:
It works fine on Linux with all three.

jdemeyer:
> It works with Python 2.7. But fails with 3.5 and 3.6.

Are you running the tests for Py2 and Py3 in a substantially different way? Are you using different scripts, a different environment, a different machine for those tests?

jdemeyer:
Is it possible to see the output of `ulimit -a` before running the tests (both on Py2 and Py3)?

jdemeyer:
In any case, the proper fix is probably to just ignore the `ValueError`. I'm just asking you those questions to confirm that.

saraedum:
The tests run on TravisCI. The machines should be virtually identical and the environment should be identical except for the versions of Python. The script is certainly the same. I'll add `ulimit -a` and report back.

saraedum:
This is for the Python 2.7 build:
```
+ ulimit -a
core file size          (blocks, -c) 0
data seg size           (kbytes, -d) unlimited
file size               (blocks, -f) unlimited
max locked memory       (kbytes, -l) unlimited
max memory size         (kbytes, -m) unlimited
open files                      (-n) 256
pipe size            (512 bytes, -p) 1
stack size              (kbytes, -s) 8192
cpu time               (seconds, -t) unlimited
max user processes              (-u) 709
virtual memory          (kbytes, -v) unlimited
```

And this is the Python 3.5 build
```
+ ulimit -a
core file size          (blocks, -c) 0
data seg size           (kbytes, -d) unlimited
file size               (blocks, -f) unlimited
max locked memory       (kbytes, -l) unlimited
max memory size         (kbytes, -m) unlimited
open files                      (-n) 256
pipe size            (512 bytes, -p) 1
[...]

PR #154, opened 2021-11-24
attempt to fix os x test suite

kliem:
The issue first fixed in #71 reappeared:

See e.g. https://github.com/sagemath/cysignals/actions/runs/1494362490

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@
 
 VERSION = $(shell cat VERSION)
 
-PYTHON = python
+PYTHON = python3
 PIP = $(PYTHON) -m pip -v
 LS_R = ls -Ra1
 
--- a/Makefile
+++ b/Makefile
@@ -6,11 +6,7 @@ PYTHON = python
 PIP = $(PYTHON) -m pip -v
 LS_R = ls -Ra1
 
-# We add ulimit -s 1024 in this Makefile to work around a very strange
-# OS X bug manifesting itself with Python 3 and old versions of GNU make.
-# This was discovered at https://github.com/sagemath/cysignals/issues/71
-# but I don't know a good analysis nor solution.
-DOCTEST = ulimit 2>/dev/null -s 1024; $(PYTHON) -B rundoctests.py
+DOCTEST = $(PYTHON) -B rundoctests.py
 
 
 #####################
--- a/example/setup.py
+++ b/example/setup.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 from setuptools import setup
 from distutils.command.build import build as _build
--- a/rundoctests.py
+++ b/rundoctests.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 #
 # Run doctests for cysignals
 #
--- a/rundoctests.py
+++ b/rundoctests.py
@@ -61,7 +61,11 @@ if os.name != 'nt':
     import resource
     # Limit stack size to avoid errors in stack overflow doctest
     stacksize = 1 << 20
-    resource.setrlimit(resource.RLIMIT_STACK, (stacksize, stacksize))
+    if sys.platform != 'darwin':
+        # Work around a very strange OS X.
+        # This was discovered at https://github.com/sagemath/cysignals/issues/71.
+        # The original solution did not last very long and the issue reappeared.
+        resource.setrlimit(resource.RLIMIT_STACK, (stacksize, stacksize))
 
     # Disable core dumps
     resource.setrlimit(resource.RLIMIT_CORE, (0, 0))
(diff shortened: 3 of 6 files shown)


dimpase:
Does it still suppport py2?
I had to do
```diff
--- a/rundoctests.py
+++ b/rundoctests.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 #
 # Run doctests for cysignals
 #
```
in order to
```
% ./rundoctests.py  
Doctesting 0 files.
  File "/Users/dima/software/cysignals/./rundoctests.py", line 64, in <module>
    resource.setrlimit(resource.RLIMIT_STACK, (stacksize, stacksize))
ValueError: current limit exceeds maximum limit
% uname -a
Darwin oucl13243.lan 21.1.0 Darwin Kernel Version 21.1.0: Wed Oct 13 17:33:23 PDT 2021; root:xnu-8019.41.5~1/RELEASE_X86_64 x86_64
```

kliem:
It assumes that $PYTHON is set to your python. So you may run `make ... PYTHON=python3`.

dimpase:
This seems to be relevant here:
https://bugs.python.org/issue34602
https://bugs.python.org/issue40518

dimpase:
Does `$PYTHON` override `#!/usr/bin/env python`? This would be weird...

Anyhow, what's the correct way to run the tests?

kliem:
The difference to other peoples problems is the following:

    # Limit stack size to avoid errors in stack overflow doctest
    stacksize = 1 << 20
    resource.setrlimit(resource.RLIMIT_STACK, (stacksize, stacksize))

So we want to limit the stack size instead of increasing it. So I guess if for some reason macos tells us `1 << 20` is too large, we just ignore that.

I really don't know what that shebang line is supposed to do. I can't tell that it makes a difference for me.

dimpase:
Indeed, I suppose one should not call that `setrlimit` on macOS at all.

Meanwhile, I really need 
```
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
on setup.py, otherwise make fails as it tries Python2, no matter if I export PYTHON=python3 or not.
What works is `make PYTHON=python3` - probably it runs a mix of python 2 and 3 then, yuck...

It doesn't work with python2 anyway, though.

dimpase:
With the following patch, tests run:
```diff
--- a/rundoctests.py
+++ b/rundoctests.py
@@ -61,7 +61,8 @@ if os.name != 'nt':
     import resource
     # Limit stack size to avoid errors in stack overflow doctest
     stacksize = 1 << 20
-    resource.setrlimit(resource.RLIMIT_STACK, (stacksize, stacksize))
+    if sys.platform != 'darwin':
+        resource.setrlimit(resource.RLIMIT_STACK, (stacksize, stacksize))
 
     # Disable core dumps
     resource.setrlimit(resource.RLIMIT_CORE, (0, 0))
```

dimpase:
here is what I get if PYTHON=python3 is not passed to make:
```
% make 
python setup.py build
running build
running build_py
running configure
  File "setup.py", line 217, in <module>
    bdist_egg=no_egg
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/Extras/lib/python/setuptools/__init__.py", line 145, in setup
    return distutils.core.setup(**attrs)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/command/build.py", line 127, in run
    self.run_command(cmd_name)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/cmd.py", line 326, in run_command
    self.distribution.run_command(command)
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "setup.py", line 173, in run
    super().run()
TypeError: super() takes at least 1 argument (0 given)
make: *** [build] Error 1
```

Seems like a bug - noone tested this on py2 for ages it seems.

dimpase:
OK, the tests on your branch pass, and thgey have https://github.com/sagemath/cysignals/pull/154#issuecomment-977776188 in, so good to go.

kliem:
Should we modify the shebang line yet?

It seems for me that `make check-all` does not work no matter what. I need to run `make check-all PYHON=python3`.

Btw, support for python < 3.6 has been dropped.

dimpase:
> Should we modify the shebang line yet?

if we're droppying Python 2 officially, then why not. By the way, Makefile has `PYTHON = python` at the top, no wonder
it's broken if good PYTHON is not passed. Not all systems set `python` to anything nowadays, you need `python3` (or `python2`, but do we care about python2 still?)

PR #163, opened 2022-01-21
Signal safety

tornaria:
Follow up to #162.

The first commit gets rid of all usage of stdio in signal handlers (mostly, but not only, inside `ENABLE_DEBUG_CYSIGNALS`).

The second commit replaces use of `gettimeofday()` by `clock_gettime()`, the latter being safe according to `signal-safety(7)`. It's also monotonic, unlike `gettimeofday()`.

Other stuff that happens inside the signal handler:

 - `cysigs_interrupt_handler()` calls `PyErr_SetInterrupt()` (probably ok?).

 - when `debug_level >= 2` the function `do_raise_exception()` calls PyGILState_Ensure / PyErr_Occurred / PyGILState_Release, I wonder if this could deadlock for some reason; in any case I don't understand why the GIL is necessary if the return value of PyErr_Occurred() is not used, just printed.

 - function `sigdie()` uses `getenv()` which is not safe, maybe the env variables should be read by `init_cysignals()` instead.

 - function `print_enhanced_backtrace()` uses `execvp()`, this is probably ok although the "correct" way would be to read PATH on init and here use `execv()` with the full path to `cysignals-CSI` already precomputed.

 - the function `sig_raise_exception()` is in cython and calls lots of python stuff, I wonder if this should be called *after* the longjmp() instead (it probably makes no difference since I don't think `longjmp()` will fix internal state of e.g. stdio).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -27,7 +27,7 @@ CXX="$CC"
 CXXFLAGS="$CFLAGS"
 AC_PROG_CXX()
 
-AC_CHECK_HEADERS([execinfo.h sys/mman.h sys/prctl.h sys/time.h sys/wait.h windows.h])
+AC_CHECK_HEADERS([execinfo.h sys/mman.h sys/prctl.h time.h sys/wait.h windows.h])
 AC_CHECK_FUNCS([fork kill sigprocmask sigaltstack backtrace])
 
 
--- a/src/cysignals/implementation.c
+++ b/src/cysignals/implementation.c
@@ -155,6 +155,52 @@ static inline void reset_CPU(void)
 #endif
 }
 
+static inline void print_stderr(const char* s)
+{
+    /* Using stdio (fputs, fprintf, fflush) from inside a signal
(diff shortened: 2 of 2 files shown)


tornaria:
> @malb @dimpase should this be merged? It seems to pass all tests. @tornaria Are you using these fixes in production already?

I'm only using #162 (see https://github.com/void-linux/void-packages/tree/master/srcpkgs/python3-cysignals/patches). This is the only issue that I know how to reproduce. This PR "fixes" potential issues of the same kind that I don't know how to reproduce.

Note that #162 is merged but never released. It would be nice to have a new release which fixes this issue and fully supports cython 3. The easy way to support cython 3 is to use `legacy_implicit_noexcept` but a proper fix is to actually add all the `noexcept` clauses so the code builds fine with cython 0.29 or cython 3.

dimpase:
rebased over the latest main, to see how it goes with CI

PR #166, opened 2022-05-30
Remove PARI

### LINKED ISSUE #125: Replace compile-time dependency on pari by run-time dispatch

kliem:
As a follow-up on the merged PR
- #181 

... this PR removes the optional-compile dependency on pari as proposed in the discussion of https://github.com/sagemath/cypari2/pull/109.

This should be merged only after 
- [X] a Cysignals release 1.11.x has been made that includes #181 
- [ ] a CyPari2 release has been made that uses the general signal hook added in #181. --> 2.2.0

Resolves #125

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -31,7 +31,6 @@ Requirements
 - Python >= 3.6
 - Cython >= 0.28
 - Sphinx >= 1.6 (for building the documentation)
-- PARI/GP (optional; for interfacing with the PARI/GP signal handler)
 
 Links
 -----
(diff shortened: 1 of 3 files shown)


videlec:
Sounds good. Though this is lacking documentation!

kliem:
I added some documentation as requested.

mkoeppe:
Is this waiting for further review, or what's the status here?

PR #174, opened 2023-08-04
Cython complaints about place of nogil

malb:

PR #176, opened 2023-09-18
Fix Cython 3 complaints about placement of nogil

mkoeppe:
Split out from #174 as suggested in https://github.com/sagemath/cysignals/pull/174#issuecomment-1713004955


malb:
lgtm

PR #181, opened 2023-09-21
Add general signal hook but keep pari

mkoeppe:
Introduces a general hook as proposed in the discussion of https://github.com/sagemath/cypari2/pull/109.

Split out from #166. @kliem @videlec

### DIFF
--- a/src/cysignals/implementation.c
+++ b/src/cysignals/implementation.c
@@ -62,6 +62,37 @@ static int PARI_SIGINT_block = 0;
 static int PARI_SIGINT_pending = 0;
 #define paricfg_version NULL
 #endif
+
+// Custom signal handling of other packages.
+#define MAX_N_CUSTOM_HANDLERS 16
+
+static int (*custom_signal_is_blocked_pts[MAX_N_CUSTOM_HANDLERS])();
+static void (*custom_signal_unblock_pts[MAX_N_CUSTOM_HANDLERS])();
+static void (*custom_set_pending_signal_pts[MAX_N_CUSTOM_HANDLERS])(int);
+static int n_custom_handlers = 0;
+
+int custom_signal_is_blocked(){
+    // Check if a custom block is set.
+    for(int i = 0; i < n_custom_handlers; i++){
+        if (custom_signal_is_blocked_pts[i]())
+            return 1;
+    }
+    return 0;
+}
+
+void custom_signal_unblock(){
+    // Unset all custom blocks.
+    for(int i = 0; i < n_custom_handlers; i++)
+        custom_signal_unblock_pts[i]();
+}
+
+
+void custom_set_pending_signal(int sig){
+    // Set a pending signal to custom handlers.
+    for(int i = 0; i < n_custom_handlers; i++)
+        custom_set_pending_signal_pts[i](sig);
+}
+
 #if HAVE_WINDOWS_H
 /* We must include <windows.h> after <pari.h>
  * See https://github.com/sagemath/cysignals/issues/107 */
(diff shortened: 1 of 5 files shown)


mkoeppe:
@malb @dimpase Let's merge this please

dimpase (commented):
is it something untested?

mkoeppe:
Yes, this has already been tested, and the corresponding PR https://github.com/sagemath/cypari2/pull/117 was already merged (.... by you, @dimpase) into cypari2, until someone decided to back it out again without discussion. https://github.com/sagemath/cypari2/pull/130 has the change on the cypari2 side of things again.

Here I am using an approach that needs less coordination. Keeping the cypari2-specific code (so that current cypari2 works) and adding a general hook (so that future cypari2 with https://github.com/sagemath/cypari2/pull/130 can work).

src/cysignals/implementation.c
```diff
@@ -400,6 +432,7 @@ static void _sig_on_interrupt_received(void)
     cysigs.sig_on_count = 0;
     cysigs.interrupt_received = 0;
     PARI_SIGINT_pending = 0;
+    custom_signal_unblock();
```
tornaria, line 435: I think there is a typo in this line: it should have been
```
    custom_set_pending_signal(0);
```
instead.

In fact, I was having some issues with cysignals 1.12.2 (some doctest failures having to do with alarms, etc) and changing this seems to sovle them.


@kliem can you confirm this is correct?



@antonio-rojas I think this is the fix we were missing. I'm testing, will do a PR tomorrow. Because of the `PARI_SIGINT_pending = 0;` line, the typo doesn't cause trouble (i.e., until that line is removed).

To be clear, this is the patch I'm testing on top of 1.12.2:
```diff
--- a/src/cysignals/implementation.c
+++ b/src/cysignals/implementation.c
@@ -591,7 +591,7 @@ static void _sig_on_interrupt_received(void)
     do_raise_exception(cysigs.interrupt_received);
     cysigs.sig_on_count = 0;
     cysigs.interrupt_received = 0;
-    custom_signal_unblock();
+    custom_set_pending_signal(0);
 
 #if HAVE_SIGPROCMASK
     sigprocmask(SIG_SETMASK, &oldset, NULL);
```

antonio-rojas: Doesn't quite work for me. Now instead of crashing, tests hang and timeout after 30 minutes.

dimpase: I've released cypari2 2.2.2

antonio-rojas: ok, not sure what happened before (possibly some memory issues), but it seems to be working now, except that the random failure I reported in https://github.com/sagemath/sage/pull/38749#issuecomment-2458913173 is now permanent (regardless of the used seed)

tornaria: > ok, not sure what happened before (possibly some memory issues), but it seems to be working now, except that the random failure I reported in [sagemath/sage#38749 (comment)](https://github.com/sagemath/sage/pull/38749#issuecomment-2458913173) is now permanent (regardless of the used seed)

Great! I have a PR almost ready to submit (including a test to trigger this bug with cypari2).

As for the remaining issue, I'm getting the same, and I'm working on it. This appears to be related to https://github.com/sagemath/cysignals/issues/215. The implementation of `sig_occurred()` seems to be incorrect, in the sense that `sig_occurred()` will sometimes return an old exception that has been already handled.

dimpase: is it a basically a 1-line PR (+ the test)?

tornaria: > is it a basically a 1-line PR (+ the test)?

Yes, I wanted to test it further. I'm also working on the remaining issue which seems more tricky, but I think I have a reasonable fix. Hopefully this might get rid of all the stupid heisenbugs that have been pestering us.

I was actually trying to push this to GH a moment ago, but it seems there is a major outage right now (git pull/push from/to github, are not working atm).

PR #183, opened 2023-09-21
.github/workflows/ci-sage.yml: Remove defunct Sage/Cygwin integration test

mkoeppe:
It's failing and would have to be replaced by using the reusable Cygwin workflow from Sage -- but that's broken too.


malb:
Happy to merge it, @dimpase wdyt?

PR #184, opened 2023-09-21
.github/workflows/wheels.yml: Deploy sdist and wheels to PyPI 

mkoeppe:
Copied from https://github.com/sagemath/sage/blob/develop/.github/workflows/dist.yml


mkoeppe:
@malb @dimpase Please review and merge

PR #187, opened 2023-10-04
make docbuild more robust, add noexcept's

dimpase:
currently make doc fails; this is a fix that
runs sphinx on already installed package.


mkoeppe (approved):
LGTM

PR #189, opened 2023-10-07
Include generated `configure` script in sdist

mkoeppe:
Our 1.11.3 sdist on PyPI, generated for the first time by a GH Actions run, does not include the `configure` script.
See https://github.com/sagemath/sage/pull/36402#issuecomment-1751773825


mkoeppe:
@dimpase @malb Please review, merge & tag

dimpase (approved):
So picking other targets for make makes sure ./configure is there?

OK, I presume you checked.

PR #204, opened 2024-07-06
Install cysignals-CSI-helper as package data for better portability

### LINKED ISSUE #200: Use importlib.resources to find cysignals-CSI-helper.py
Currently `cysignals-CSI-helper.py` is installed  https://github.com/sagemath/cysignals/blob/main/setup.py#L219. The script that makes use of it, `cysignals-CSI`, later tries to guess the install location based on `sys.argv[0]`: https://github.com/sagemath/cysignals/blob/main/src/scripts/cysignals-CSI#L68

This can fail if `sys.argv[0]` isn't something simple like `/usr/bin/python`, e.g. https://bugs.gentoo.org/927767

By using `importlib.resources.files` we could avoid guessing at all.

mgorny:
Rather than installing `cysignals-CSI-helper.py` into a `share` directory and then trying to figure out the correct path to it, install it as Python package data and use the standard `importlib.resources` API to access it.  For Python versions older than 3.9, the `importlib_resources` backport is used instead.

Fixes #200


orlitzky:
Thanks for working on this.

orlitzky:
@mkoeppe any reason not to merge this? Without it, we guess the wrong path on Gentoo.

dimpase:
This should be updated, as we don't deal with python < 3.9 here any more. @tobiasdiez - does this fit into the meson setup?

tobiasdiez:
Yes, one can easily control where the script should be installed.
https://github.com/sagemath/cysignals/blob/main/src/scripts/meson.build

dimpase:
@mgorny - please take a look at the PR to move this forward: https://github.com/mgorny/cysignals/pull/1

mgorny:
Err, that PR doesn't apply cleanly to my branch, i.e. GH doesn't let me merge it. Feel free to just overwrite my branch (it's open to maintainers) or replace the PR.

dimpase:
> Err, that PR doesn't apply cleanly to my branch, i.e. GH doesn't let me merge it. Feel free to just overwrite my branch (it's open to maintainers) or replace the PR.

while maintainers are too busy, and I'm not one, the quickest way forward would be just to force-push my branch into your PR.

mgorny:
> > Err, that PR doesn't apply cleanly to my branch, i.e. GH doesn't let me merge it. Feel free to just overwrite my branch (it's open to maintainers) or replace the PR.
> 
> while maintainers are too busy, and I'm not one, the quickest way forward would be just to force-push my branch into your PR.

Done.

dimpase:
@vbraun - can you promote me to maintainer here, or otherwise take care of this and other PRs waiting here?

dimpase:
I haven't figured out to invoke `testgdb.py` from `pytest`. Any pointers?

tobiasdiez:
> I haven't figured out to invoke `testgdb.py` from `pytest`. Any pointers?

This still uses `unittest`. So one needs to first migrate it to pytest, and then rename it to end on `_test.py` so that it gets automatically collected by pytest. I'll open an issue to track this.

Issue #179, opened 2023-09-18
`configure` tests the wrong compiler


mkoeppe:
Extension modules are built with the compiler configured in distutils/setuptools. 
When `setup.py` calls the configure script, it should pass this compiler.

This will eliminate this error seen in https://github.com/sagemath/cysignals/actions/runs/6227465351/job/16902173349:
```
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether the compiler supports GNU C... yes
checking whether gcc accepts -g... yes
checking for gcc option to enable C11 features... none needed
checking whether the compiler supports GNU C++... yes
checking whether gcc accepts -g... yes
checking for gcc option to enable C++11 features... none needed
[...]
checking for emms instruction... yes

 clang -Wsign-compare -Wunreachable-code -fno-common -dynamic -DNDEBUG -g -fwrapv -O3 -Wall -arch arm64 -arch x86_64 -g -DCYTHON_CLINE_IN_TRACEBACK=0 -U_FORTIFY_SOURCE -Isrc/cysignals -Isrc -I/Library/Frameworks/Python.framework/Versions/3.11/include/python3.11 -c build/src/cysignals/signals.c -o build/temp.macosx-10.9-universal2-cpython-311/build/src/cysignals/signals.o -pthread
  In file included from build/src/cysignals/signals.c:1728:
  build/src/cysignals/implementation.c:539:13: warning: unused function 'print_enhanced_backtrace' [-Wunused-function]
  static void print_enhanced_backtrace(void)
              ^
  build/src/cysignals/implementation.c:123:9: error: unrecognized instruction mnemonic
      asm("emms");
          ^
  <inline asm>:1:2: note: instantiated into assembly here
          emms
          ^
  1 warning and 1 error generated.
  error: command '/usr/bin/clang' failed with exit code 1
  error: subprocess-exited-with-error
  
  Building wheel for cysignals (pyproject.toml): finished with status 'error'
  × Building wheel for cysignals (pyproject.toml) did not run successfully.
Failed to build cysignals
  │ exit code: 1
  ╰─> See above for output.
```

mkoeppe:
Specifically, Python attempts to build using `-arch arm64 -arch x86_64` -- but configure only tested for one of the two archs.

mkoeppe:
[...]

PR #206, opened 2024-10-29
Update build system to use Meson

tobiasdiez:
Following the example of Scipy/Numpy and https://github.com/sagemath/sage/pull/36524, the build system is migrated to Meson.

I tried to stay as close as possible to the original autoconf implementation. A few differences are:
- No longer need to auto-generate `signals.pxd`, but have it at as a normal source file.
- Changed the authors to `Sage Developers` to align with the main sage repo. I'm unsure about this one and happy to revert it.
- Drop Python < 3.9 in alignment with https://scientific-python.org/specs/spec-0000/
- Remove the special handling when `MINSIGSTKSZ_IS_CONSTANT` (I had trouble making it compile and it's not constant for recent versions of glibc anyway)
- Remove the "ci-with-sage" jobs in the github action, which are non-working on the main branch and I've also no idea how to make them compatible with meson; I also think the current ci runs should give a pretty good coverage.
- Include the code of `implementation_cygwin.c` directly in `implementation.c` since I was not able to make it compile (it's also a bit hard for me to work on the cygwin code since I only have the ci to test it.)

Closes https://github.com/sagemath/cysignals/issues/179 (obsolete now)


tobiasdiez:
Tests are green, so this is also ready to go.

Please squash after https://github.com/sagemath/cysignals/pull/209.

dimpase:
something happened during a rebase? There are no `meson.build` files anywhere

dimpase (changes requested):
please check that the rebase went OK (it seems it hasn't)

tobiasdiez:
Right, thanks. Something went wrong with merging https://github.com/sagemath/cysignals/pull/209 into this branch here. I've kinda fixed it locally, but it would be easier if https://github.com/sagemath/cysignals/pull/209 can be merged first (and squashed!) into main.

tobiasdiez:
Uff, tests are finally green.

fchapoton:
Would you agree to engage yourself to maintain this for the years to come ?

dimpase:
meson is a widely used and documented, unlike our present system to build sagelib.

fchapoton:
So we both agree that it's a good idea to merge, right ?

tobiasdiez:
> Would you agree to engage yourself to maintain this for the years to come ?

That's hard to promise but yes I'm willing to (help) maintaining it. The meson build info is also pretty straightforward, so it shouldn't be too hard for someone else to modify it.

Please merge #209 before this one here, and squash this PR to have a cleaner git history. Thanks!

dimpase (approved):
lgtm

tobiasdiez:
Thanks @dimpase and @fchapoton!

Issue #212, opened 2024-11-25
Consumers failing to build with v1.12.0


orlitzky:
We just added v1.12.0 in Gentoo and switched the build to meson but now the packages that depend on it are failing to build: fpylll, primecountpy, cypari, pplpy, etc.

A typical error:

> 2024-11-22 07:13:20,452 root INFO building 'fpylll.fplll.wrapper' extension
> 2024-11-22 07:13:20,454 root INFO x86_64-pc-linux-gnu-g++ -O2 -pipe -march=native -fno-diagnostics-color -DNDEBUG -fPIC -Isrc/fpylll/fplll -I/usr/lib/python3.12/site-packages/cysignals -I/usr/include -I/usr/lib/python3.12/site-packages/numpy/_core/include -I/usr/include/python3.12 -c build/src/fpyll
> In file included from build/src/fpylll/fplll/lll.cpp:1353:
> /usr/lib/python3.12/site-packages/cysignals/macros.h: In function int _sig_on_prejmp(const char*, const char*, int):
> /usr/lib/python3.12/site-packages/cysignals/macros.h:117:5: error: cysigs was not declared in this scope
>   117 |     cysigs.s = message;
>       |     ^~~~~~

I tried to debug for a few minutes tonight but it will probably have to wait until next week.

ludopulles:
Actually, already compiling the [basic example](https://cysignals.readthedocs.io/en/latest/interrupt.html#basic-example) with `cythonize -i FILE.pyx` gives me those errors (cython version 3.0.11).

It seems that the meson build system doesn't put `#include "macros.h"` after the late_includes `/* #### Code section: late_includes ### */`.

antonio-rojas:
Caused by the removal of the line `# cython: preliminary_late_includes_cy28=True` from `signals.pxd` in https://github.com/sagemath/cysignals/commit/fa6d45d2ac1f88d45583a8431ddee811fb2bd617

antonio-rojas:
> @antonio-rojas: Do these `# cython` directives in headers carry over to downstream projects?

Yes, adding that line back fixes the build issues in dependant projects

> If that's indeed the core issue, then a workaround for other projects would be to compile using `-X preliminary_late_includes_cy28=True`.

But why should other projects work around this when it's the cysignals header itself that needs it?

ludopulles:
I agree with Antonio it's an issue within the `cysignals` package, not the dependent projects.
[...]

PR #213, opened 2024-12-06
Fix build issues

ludopulles:
The migration to the meson build system, causes failures when building packages that depend on `cysignals` but don't use meson (i.e. they use `setup.py`).

Closes: #212

### DIFF
--- a/example/meson.build
+++ b/example/meson.build
@@ -15,13 +15,6 @@ print(cysignals.__file__.replace('__init__.py', ''))
 ).stdout().strip()
 cysignals = declare_dependency(include_directories: inc_cysignals)
 
-# Make declarations in Cython code available to C include files
-add_project_arguments(
-  '-X preliminary_late_includes_cy28=True',
-  language: 'cython',
-)
-
-
 py.extension_module('cysignals_example',
   sources: ['cysignals_example.pyx'],
   install: true,
--- a/meson.build
+++ b/meson.build
@@ -22,11 +22,6 @@ add_project_arguments('-DCYTHON_CLINE_IN_TRACEBACK=0', language: 'c')
 # Disable sanity checking in GNU libc
 # This is required because of false positives in the longjmp() check
 add_project_arguments('-U_FORTIFY_SOURCE', language: 'c')
-# Make declarations in Cython code available to C include files
-add_project_arguments(
-  '-X preliminary_late_includes_cy28=True',
-  language: 'cython',
-)
 
 # Platform-specific settings
 if is_cygwin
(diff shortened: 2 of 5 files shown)


tobiasdiez:
@fchapoton @dimpase could you please have a look and merge this (and perhaps directly cut a new release).

I'm very sorry for having introduced this bug during the meson migration.

dimpase:
I don't understand why this PR touches meson.build files. IMHO removing
```
add_project_arguments(
  '-X preliminary_late_includes_cy28=True',
  language: 'cython',
```
there might create meson build problems.

example/meson.build
```diff
@@ -15,13 +15,6 @@ print(cysignals.__file__.replace('__init__.py', ''))
 ).stdout().strip()
 cysignals = declare_dependency(include_directories: inc_cysignals)
 
-# Make declarations in Cython code available to C include files
```
dimpase, line 18: As far as I know, meson does not understand `# cython:` directives.  Anyway, there is no need to touch `meson.build` files in this PR.

dimpase (changes requested):

example/meson.build
tobiasdiez: I also thought so but Cython actually still looks for these magic comments. So it's okay to remove change the meson files. What is definitely no longer working are `#distutils` comments.

dimpase:
OK, good then

example/meson.build
user202729: If you think about it it makes sense — you need to give meson precisely the information needed for it to know which commands should be run etc.

So e.g. `language=c++` is needed, so that meson knows it should run `gcc a.cpp` instead of `gcc a.c`. Dependencies specification is needed so that meson knows it should rebuild the file when dependency version is updated.

The late include flag… probably not so much.

ludopulles: (for future reference:)

https://cython.readthedocs.io/en/latest/src/userguide/source_files_and_compilation.html#globally

The cython compiler takes the directives straight from the pyx/pxd and/or command line. Because it's now placed in the file, you don't have to pass it (via meson) to the `cython` command (although meson *can* override compiler directives in the file by adding project arguments).

ludopulles:
Thanks for merging this :smile:.
Is there also a new release planned in the near future?

dimpase:
hopefully later today

user202729:
@dimpase I see you published the new release on GitHub, but you didn't update version number in pyproject.toml?

dimpase:
oops. OK, I'll do another one, also including one more just merged PR, by you. 

Build systems ought to get the version from the git tag...

PR #217, opened 2025-01-14
Fix warnings (C and cython)

tornaria:
This PR fixes warnings when building and testing

- **avoid C warnings**
- **cython warnings: disable performance hints in tests.pyx**
- **cython warnings: enable -Wextra**
- **cython warnings: avoid unused variables**
- **cython warnings: avoid maybe uninitialized**
- **reload cysignals handlers in tests.pyx**
- **tests::c_test_sig_on_cython: should not be noexcept**
- **alarm::setitimer_real: doesn't raise exceptions**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,6 +39,6 @@ jobs:
           pip install --upgrade -r requirements.txt
           pip install --upgrade ninja
     - name: Build
-      run: pip install --no-build-isolation --config-settings=builddir=builddir .
+      run: pip install -v --no-build-isolation --config-settings=builddir=builddir .
     - name: Test
       run: meson test --print-errorlogs -C builddir
--- a/meson.build
+++ b/meson.build
@@ -1,5 +1,5 @@
 project('cysignals', 'c', 'cpp', 'cython',
-  default_options: ['warning_level=3', 'cpp_std=c++17']
+  default_options: ['warning_level=2', 'cpp_std=c++17']
 )
 
 # Python
--- a/src/conftest.py
+++ b/src/conftest.py
@@ -4,6 +4,9 @@ import platform
 from _pytest.nodes import Collector
 from _pytest.doctest import DoctestModule
 
+from _pytest.pathlib import resolve_pkg_root_and_module_name
+import importlib
+
 # cysignals-CSI only works from  gdb, i.e. invoke ./testgdb.py directly
 collect_ignore = ["cysignals/cysignals-CSI-helper.py"]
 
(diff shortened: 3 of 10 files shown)


tornaria:
@tobiasdiez Is it possible to show the pytest output?

Right now all that is shown is:
```
Run meson test --print-errorlogs -C builddir
ninja: Entering directory `/home/runner/work/cysignals/cysignals/builddir'
ninja: no work to do.
1/2 example OK               5.31s
2/2 pytest  OK              27.17s

Ok:                 2   
Expected Fail:      0   
Fail:               0   
Unexpected Pass:    0   
Skipped:            0   
Timeout:            0   

Full log written to /home/runner/work/cysignals/cysignals/builddir/meson-logs/testlog.txt
```
I can look at the file locally, but not on CI. Since pytest output is already brief, just showing it in the output would be nice, and also gives a better progress report.

----

Also, what is the suggested workflow to build and test "in place"?

I would expect something like this:
```
meson setup builddir
ninja -C builddir
meson test -C builddir
```
but this in fact tests the cysignals installed in the system. I shouldn't be required to install before testing.

Something like
```
PYTHONPATH=$(cd builddir/src && pwd) meson test -C builddir
```
or even
```
PYTHONPATH=$(cd builddir/src && pwd) pytest
```
fails because building doesn't create a complete package in `builddir/src/cysignals`, since python files are not copied (at least `__init__.py` is necessary).

Ideally, a simple way to run `pytest` directly would be nice (because it gives much more control on what to test, is more useful for developing).

tobiasdiez:
I don't think there is an easy way to print the test output directly (https://mesonbuild.com/Unit-tests.html#test-outputs). Either we could run pytest manually or afterwards just ̀ cat` the log file in the CI.

For editable install, the simplest would be the classical `pip install -e .`.

> I shouldn't be required to install before testing.

You do need to install (eg into a local venv) since the build directory contains the compiled cython code but no python files, and it's the opposite for the source folder.

tornaria:
> I don't think there is an easy way to print the test output directly (https://mesonbuild.com/Unit-tests.html#test-outputs). Either we could run pytest manually or afterwards just ̀ cat` the log file in the CI.

For CI, cat is good. For local, is not (I want to see pytest output *while it's happening*)

> For editable install, the simplest would be the classical `pip install -e .`.

I don't trust pip, I'm always afraid it will break my system. I provide the dependencies, I only want to build cysignals and test it.

> > I shouldn't be required to install before testing.
> 
> You do need to install (eg into a local venv) since the build directory contains the compiled cython code but no python files, and it's the opposite for the source folder.

Isn't it possible that a complete package is built in the build directory?

I was hoping meson-python had something figured out for the normal build / test cycle.

tobiasdiez (approved):

tornaria:
Thanks for the review. I just pushed a couple empty lines to make flake8 happy.

tornaria:
The last commit may be unnecessary after #220.

tornaria:
> The last commit may be unnecessary after #220.

Indeed. I removed the last commit from this PR.

PR #224, opened 2025-01-23
Install cypari2 for testing

tornaria:
There's one test that depends on cypari2, let's see if this works.

(this is based on top of #217)


tornaria:
This works now. For simplicity, it only tests cypari2 on ubuntu, installing pari from the os using apt-get and cypari2 using pip.

@dimpase @tobiasdiez  do you think this is a good idea?

So far there's only one test that needs cypari2 (the one from #216).

dimpase:
+1

PR #226, opened 2025-02-09
Better fix for when CFLAGS includes -D_FORTIFY_SOURCE

tornaria:
The module signals.pyx has to be built with fortify disabled, because of
a false positive in one of the `longjmp()`, more precisely in the call
```
cylongjmp(trampoline_setup, 1)
```
which appears in `setup_trampoline()`.

Often distributions will add `-D_FORTIFY_SOURCE=2` to `CFLAGS`, and so
this has to be overridden when compiling `signals.pyx` by adding
`-U_FORTIFY_SOURCE` to the compiler arguments. The former solution,
using `add_project_arguments()` doesn't work since it will add flags
*before* the ones provided in `CFLAGS`.

Instead, add `-U_FORTIFY_SOURCE` to the arguments in the definition of
the extension module `signal`, which will add it to the command line
*after* the ones provided in `CFLAGS`. In addition, fortify is disabled
only for building `signal.pyx`.


tobiasdiez (approved):
LGTM

tornaria:
@cxzhong this is better than removing the check in `implementation.c`.

Try
```
CFLAGS="-D_FORTIFY_SOURCE -O2 -Wall" meson setup build && ninja -C build
```
to see why this is useful. With this PR, it works just fine.

tornaria:
@tobiasdiez this is still useful, I only rebased on current main.

tobiasdiez:
@dimpase could you please merge this?

PR #229, opened 2025-05-17
tomli is integrated as tomllib since Python 3.11

a-detiste:


dimpase:
the minimum Python version we support is 3.11, so there is no need to special-case anything older

a-detiste:
Ok, I change

dimpase (commented):
why no tomllib in requirements.txt ? is it a part of standard Python now?

a-detiste:
Yes, exactly. It s a copy of tomli with another name.

dimpase:
> the minimum Python version we support is 3.11, so there is no need to special-case anything older

oops, I am sorry, I messed up (confused `sagemath/cysignals` with `sagemath/sage`). cysignals should still work on  python 3.9+, as you can see in https://github.com/sagemath/cysignals/blob/main/pyproject.toml

Thus, we still need to
```
import tomli as tomllib
```
on pythons older than 3.11.

a-detiste:
Ok, previous change was rebased in oblivion, will do it again.

requirements.txt
```diff
@@ -4,5 +4,5 @@ wheel
 Cython
 Sphinx
 flake8
-tomli
+tomli; python_version < '3.11'
```
dimpase, line 7: I don't get why it says `tomli` rather than `tomlli`. (or it should be `tomli` in `conf.py`, not `tomlli`

dimpase:
Thanks!

PR #230, opened 2025-07-20
Fix the atomicity of cy_atomic_int

ptrrsn:
When building sage, `meson.build` incorrectly sets all `CYSIGNALS_*` macros to 0 because of lacks of `main` function, which leads to a non-atomic declaration of `cy_atomic_int`.

Moreover, the `cy_atomic_int` typedefs in `struct_signals.h` was structured in a way that when it is compiled by a C++ compiler in an environment with a C compiler that supports `_Atomic`, it would never use `std::atomic`, which is incorrect and might cause compilation problem when the C++ compiler does not support `_Atomic`. This problem was hidden by the incorrect configuration of `meson.build` above but got surfaced after the `meson.build` fix.

As a context: https://github.com/LLewark/khoca/pull/3

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -82,9 +82,9 @@ config.set('CYSIGNALS_USE_SIGSETJMP', (setjmp_saves_mask or gnulibc) ? 1 : 0)
 
 # Check for atomic operations
 # for _Atomic in C code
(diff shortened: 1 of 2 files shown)


orlitzky:
`int main(void)` please, for consistency across the various C standards

ptrrsn:
Thank you for the reviews and merge.

ptrrsn:
Thank you, I can see it now. cc @soehms

Issue #215, opened 2024-12-30
Nested signal handling leads to excessive slowdown of `sig_occurred()` 


user202729:
From https://github.com/sagemath/sage/issues/39224

SageMath session:

```
sage: %%cython
....: from cysignals.signals cimport sig_on, sig_off, sig_occurred
....: from libc.stdlib cimport abort
....: def f():
....:     try:
....:         sig_on()
....:         abort()
....:         sig_off()
....:     except:
....:         raise ValueError
....: def g():
....:     for i in range(50):
....:         ignore = sig_occurred()
sage: %time g()  # fast
CPU times: user 4 μs, sys: 0 ns, total: 4 μs
Wall time: 6.44 μs
sage: f()
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
File /tmp/tmpgqxu9lji/spyx/_tmp_tmpfkjflhcs_tmp_f70vs9fj_pyx/_tmp_tmpfkjflhcs_tmp_f70vs9fj_pyx_0.pyx:5, in _tmp_tmpfkjflhcs_tmp_f70vs9fj_pyx_0.f()
      4 try:
----> 5     sig_on()
      6     abort()

RuntimeError: Aborted

During handling of the above exception, another exception occurred:

ValueError                                Traceback (most recent call last)
Cell In[3], line 1
----> 1 f()

File /tmp/tmpgqxu9lji/spyx/_tmp_tmpfkjflhcs_tmp_f70vs9fj_pyx/_tmp_tmpfkjflhcs_tmp_f70vs9fj_pyx_0.pyx:9, in _tmp_tmpfkjflhcs_tmp_f70vs9fj_pyx_0.f()
      7         sig_off()
      8     except:
----> 9         raise ValueError
     10 def g():
     11     for i in range(50):

ValueError: 
sage: %time g()  # slow
CPU times: user 3.84 s, sys: 133 μs, total: 3.84 s
Wall time: 3.85 s
sage: raise RuntimeError
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In[5], line 1
----> 1 raise RuntimeError

RuntimeError: 
sage: %time g()  # fast again
CPU times: user 79.9 ms, sys: 0 ns, total: 79.9 ms
Wall time: 79.9 ms
```

Note the invocation of `g()` that is marked `# slow`.

user202729:
[...]

PR #233, opened 2025-09-19
Update Python version requirements to 3.11 and above, and add support for 3.14

tobiasdiez:
Following https://scientific-python.org/specs/spec-0000/. Could probably even consider dropping 3.11.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ['macos-13', 'macos-latest', 'ubuntu-latest', 'windows-latest']
-        python-version: ['3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.11', '3.12', '3.13', '3.14']
     steps:
     - name: Set up the repository
       uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


tobiasdiez:
> or I can just not upgrade to Python 3.14 and not encounter the issue 🙈

Or… you could upgrade and live on the wild side 😄 Anyway, I was referring more to the fact that you successfully looked at and fixed similar bugs before - so I thought you might like to look at this one as well.

 
> speaking of the most recent commit, if [sagemath/sage#40734](https://github.com/sagemath/sage/pull/40734) gets in then `sig_check` would no longer be necessary for sage itself (downstream may still use it, but `sig_check` has a few disadvantages, one is it may call `gc.collect()` so you really don't want to call it in a tight loop and **especially** not in the destructor of something that may have a few hundred instances created such as `Integer`, another is it can be unreliable in the presence of reference cycles i.e. it will err too much on the side of caution in that case, keep reporting `True` and try to `gc.collect()`, but anyway this is the topic of [#215 (comment)](https://github.com/sagemath/cysignals/issues/215#issuecomment-2565535289) )

Good to know. I'm sadly not familiar enough with the code to properly review https://github.com/sagemath/sage/pull/40734, so you might try to ping a few people that worked on it/similar code before. Once `sig_check` is no longer used in sage, we can deprecated it here.

PR #237, opened 2025-10-29
Fix release workflow by removing duplicate macos building and migrating to Trusted Publisher

tobiasdiez:
The wheel publishing during the release failed since the macos wheels were built twice, and then the `download-artificat` action produced garbage (see eg https://github.com/scikit-build/scikit-build-core/issues/696).
This is fixed by only building on the latest macos arm (we could probably also produce wheels for macos intel, but since this platform is discontinued and wheels are not super important for cysignals anyway, I don't see much value in doing this).

Moreover, I've removed the discouraged user & password auth for publishing. This means however that Trusted Publishing has to be setup:

A new Trusted Publisher for the currently running publishing workflow can be created by accessing the following link(s) while logged-in as an owner of the package(s):
- https://pypi.org/manage/project/cysignals/settings/publishing/?provider=github&owner=sagemath&repository=cysignals&workflow_filename=dist.yml

(see also https://docs.pypi.org/trusted-publishers/adding-a-publisher/)

### DIFF
--- a/.github/workflows/dist.yml
+++ b/.github/workflows/dist.yml
@@ -35,7 +35,6 @@ jobs:
           - ubuntu-24.04-arm
           - windows-latest
           - windows-11-arm
-          - macos-14
           - macos-latest
     steps:
       - uses: actions/checkout@v4
@@ -54,8 +53,6 @@ jobs:
     needs: [build_wheels, make_sdist]
     if: 
     runs-on: ubuntu-latest
-    env:
-      CAN_DEPLOY: ${{ secrets.SAGEMATH_PYPI_API_TOKEN != '' }}
     permissions:
       id-token: write
       attestations: write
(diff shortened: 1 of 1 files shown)


tobiasdiez:
Thanks, that should have been everything. If you create a new release, it should work now... (hopefully ;-))

dimpase:
well, the action "succeeds", but nothing gets published.
(I removed and re-added the 1.12.6 release on GitHub)
https://github.com/sagemath/cysignals/actions/runs/18918851682/job/54009778131

dimpase:
https://docs.pypi.org/trusted-publishers/using-a-publisher/ says that there should be an `id-token` mentioned in the action file (I can't make complete sense of how this works, but it's something that we don't set here)
PS. oops, it appears to be there. No idea what's up

tobiasdiez:
Did you push a new tag? These steps are only run for tags, not when you manually run the workflow:
https://github.com/sagemath/cysignals/blob/117f187ec8cffec57bb953ee2599efa1d4fb8e5d/.github/workflows/dist.yml#L71

dimpase:
right. I naively thought it's triggered by making a new github release, not by a new tag. All good now

tobiasdiez:
Thanks a lot @dimpase!

PR #241, opened 2026-03-11
Add ubuntu-24.04-arm to CI workflow matrix and build wheels for free-threading

### LINKED ISSUE #240: Add pre-built wheels for Linux aarch64 (manylinux + musllinux)
## Problem

cysignals 1.12.6 publishes pre-built wheels for every major platform **except Linux aarch64**:

| Platform | Wheel | Status |
|----------|-------|--------|
| macOS arm64 | `cp3*-macosx_11_0_arm64` | ✅ |
| Linux x86_64 (glibc) | `cp3*-manylinux_2_28_x86_64` | ✅ |
| Linux x86_64 (musl) | `cp3*-musllinux_1_2_x86_64` | ✅ |
| Windows x64 | `cp3*-win_amd64` | ✅ |
| Windows arm64 | `cp3*-win_arm64` | ✅ |
| **Linux aarch64 (glibc)** | — | ❌ source build |
| **Linux aarch64 (musl)** | — | ❌ source build |

Every `pip install cysignals` on Linux aarch64 triggers a full source build (Cython + C compilation), which is slow and requires a full C toolchain. This affects CI pipelines, Docker builds, and developer machines (e.g. AWS Graviton, Azure Cobalt, GCP Axion, Ampere, Hetzner CAX, Raspberry Pi).

## Root cause

Commit 117f187e ("temporary: remove ubuntu-24.04-arm from dist", 2025-10-29) removed the arm runner from the `dist.yml` matrix right before the 1.12.6 release.

However, looking at the CI run that included `ubuntu-24.04-arm` ([run #18891747780](https://github.com/sagemath/cysignals/actions/runs/18891747780)), **the arm64 wheel build succeeded**. All 7 build jobs passed — the only failure was the "Upload wheels to PyPI" step, which was a Trusted Publisher configuration issue (fixed separately in #237).

The arm64 removal appears to have been an overly cautious reaction to the upload failure, not an actual build problem.

## Suggested fix

Re-add `ubuntu-24.04-arm` to the `dist.yml` CI matrix:

```yaml
matrix:
  os:
    - ubuntu-latest        # x86_64
    - ubuntu-24.04-arm     # aarch64 (native runner)
    - windows-latest
    - windows-11-arm
    - macos-latest
```

GitHub provides [native ARM64 runners](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories/) for free on public repositories since January 2025. Native builds avoid the QEMU signal delivery issues described in #159.

## Related issues and PRs

| # | Title | Relevance |
|---|-------|-----------|
| #159 | PSelecter test fails on aarch64 | Test failure under **QEMU** (not native arm64) — should not affect native `ubuntu-24.04-arm` runners |
| #192 | fix cross-compilation | Merged fix enabling cross-compilation for arm architectures (Void Linux) |
| #223 | A few improvements in CI | Maintainer @tornaria asked: *"I wonder if we should add `ubuntu-24.04-arm` to the matrix so we test linux arm64?"* — never followed up for `dist.yml` |
| #234 | Use `uv` and simplify wheel building | Simplified wheel building with cibuildwheel — this PR **added** `ubuntu-24.04-arm` to `dist.yml` |
| #237 | Fix release workflow / Trusted Publisher | Fixed the PyPI upload failure that caused the 1.12.6 release to fail — the arm64 build itself was fine |
| #153 | Publish wheels to PyPI | Original request for pre-built wheels |
| #21 | cysignals is not thread-safe | Relevant for free-threaded Python (cp314t) but orthogonal to this platform coverage issue |

## Impact

Linux aarch64 server adoption is accelerating across all major clouds. Pre-built wheels would eliminate unnecessary source builds for all users on this platform.

cxzhong:
Fix #240 I also lint the meson.build file


.github/workflows/dist.yml
```diff
@@ -35,6 +35,12 @@ jobs:
           - windows-latest
           - windows-11-arm
           - macos-latest
+          - ubuntu-24.04-arm
+          - macos-15-intel
+
+    env:
+      CIBW_ENABLE: "cpython-freethreading"
```
tobiasdiez: please move this to the pyproject.toml (so that local runs also test it)

tobiasdiez (approved):
I don't have time for a proper review atm (eg test the changes), but the code changes look reasonable to me.

tornaria:
Maybe split the lint step to a later PR, document there how you run the linter, and perhaps add a lint check to CI? Note that @tobiasdiez reviewed it before the lint step. Also, I prefer rebase than merges, as it makes things easier to review.

It's a matter of taste, but I'd rather have PRs that do one thing. The PR itself is documentation for the future so be explicit about everything you do in the PR. Your commit message for bc916133 is already much more detailed, at the very least you should copy this to the PR message. Some changes seem ok to sneak in, but some changes maybe need to be more explicit, e.g. minimum version requirements for python and cython (python is most likely ok, although is it necessary? otoh, cython 3.1 is less than one year old).

From your log message "Fix compatibility issues with Python 3.14" can you elaborate?

cxzhong:
the minimum python requirement is sepc000. and the minimum cython requirement is since only cython>3.1.0, we can use free threading directive.

cxzhong:
Maybe I will going to 2-3 PRs about this

tornaria:
> the minimum python requirement is sepc000. and the minimum cython requirement is since only cython>3.1.0, we can use free threading directive.

But it should still be possible to build the package with older cython (for a non free-threading python)?

cxzhong:
> > the minimum python requirement is sepc000. and the minimum cython requirement is since only cython>3.1.0, we can use free threading directive.
> 
> But it should still be possible to build the package with older cython (for a non free-threading python)?

I have added `# cython: freethreading_compatible = True` in the top of files, this directive can not be recongnized for cython<3.1. It is added in cython 3.1

PR #244, opened 2026-04-04
Remove preliminary_late_includes_cy28 directive

cxzhong:
Move all static inline functions from macros.h to implementation.c and replace them with #define macros. Since macros are text templates that expand at the call site (where Cython's capsule declarations are available), this eliminates the need for late includes entirely.

The capsule mechanism shares the implementation.c functions across all importing modules without requiring any linking — this works on all platforms (Linux, macOS, Windows).

Changes:
- macros.h: Remove all static inline functions; keep only #define macros (sig_on, sig_str, sig_off, sig_check, sig_block, sig_unblock, sig_retry, sig_error)
- implementation.c: Add former inline functions as static functions (_sig_on_prejmp, _sig_on_postjmp, _sig_off_, _sig_unblock_, _sig_retry_, _sig_error_, _set_debug_level)
- signals.pxd: Remove directive; add new functions to capsule block
- signals.pyx: Remove directive; add new function declarations
- tests.pyx: Remove directive
- tests_helper.c: Remove direct cysigs reference (not needed in fork child process)

### DIFF
--- a/src/cysignals/implementation.c
+++ b/src/cysignals/implementation.c
@@ -852,7 +852,7 @@ dienow:
 /*
  * Set message, return 0 if we need to cysetjmp(), return 1 otherwise.
  */
-static int _sig_on_prejmp(const char* message, CYTHON_UNUSED const char* file, CYTHON_UNUSED int line)
+static inline int _sig_on_prejmp(const char* message, CYTHON_UNUSED const char* file, CYTHON_UNUSED int line)
 {
     cysigs.s = message;
 #if ENABLE_DEBUG_CYSIGNALS
@@ -884,7 +884,7 @@ static int _sig_on_prejmp(const char* message, CYTHON_UNUSED const char* file, C
  * Process the return value of cysetjmp().
  * Return 0 if there was an exception, 1 otherwise.
  */
-static int _sig_on_postjmp(int jmpret)
+static inline int _sig_on_postjmp(int jmpret)
 {
     if (unlikely(jmpret > 0))
     {
@@ -913,7 +913,7 @@ static int _sig_on_postjmp(int jmpret)
 /*
  * Implementation of sig_off().
  */
-static void _sig_off_(const char* file, int line)
+static inline void _sig_off_(const char* file, int line)
 {
 #if ENABLE_DEBUG_CYSIGNALS
     if (cysigs.debug_level >= 4)
@@ -934,7 +934,7 @@ static void _sig_off_(const char* file, int line)
 }
 
 
-static void _sig_unblock_(void)
+static inline void _sig_unblock_(void)
 {
 #if ENABLE_DEBUG_CYSIGNALS
     if (cysigs.block_sigint < 1)
@@ -976,7 +976,7 @@ static void _sig_error_(void)
 }
 
 
-static int _set_debug_level(int level)
+static inline int _set_debug_level(int level)
 {
 #if ENABLE_DEBUG_CYSIGNALS
     int old = cysigs.debug_level;
--- a/src/cysignals/implementation.c
+++ b/src/cysignals/implementation.c
@@ -29,6 +29,7 @@ Interrupt and signal handling for Cython
 
 
 #include "config.h"
+#include <Python.h>
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
@@ -52,7 +53,6 @@ Interrupt and signal handling for Cython
 #if HAVE_SYS_PRCTL_H
 #include <sys/prctl.h>
 #endif
-#include <Python.h>
 
 // Custom signal handling of other packages.
 #define MAX_N_CUSTOM_HANDLERS 16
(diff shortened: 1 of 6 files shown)


cxzhong:
Now we do not need to use cython0.28 compatible mode for cysignals

src/cysignals/implementation.c
```diff
@@ -160,7 +155,7 @@ static inline void print_stderr(const char* s)
 /* str should have enough space allocated */
 static inline void ulong_to_str(unsigned long val, char *str, int base)
 {
-    const char xdigits[16] = "0123456789abcdef";
+    const char xdigits[17] = "0123456789abcdef";
```
tornaria: why?

cxzhong: Since it ends with /0. the null hidden char

cxzhong: > why?

If you want to write 16bit char, you should write as 'a' 'b' 'c' like this.

tornaria:
Can you explain the issue this solves. Is it possible to solve this without repeating code and without macros?

How is it that `cdef extern from "macros.h": ...` doesn't include `macros.h` before using the functions defined in there?

src/cysignals/implementation.c
tornaria: No, the null character is only included if there is room in the array. We don't need it here.

cxzhong: but the compiler thinks the char is 17 not 16, so it raises a warning.

src/cysignals/implementation.c
```diff
@@ -23,12 +23,8 @@ Interrupt and signal handling for Cython
  ****************************************************************************/
 
 
-#if __USE_FORTIFY_LEVEL
-#error "cysignals must be compiled without _FORTIFY_SOURCE"
```
tornaria, line 27: Is it now safe to build with `_FORTIFY_SOURCE`?

src/cysignals/implementation.c
tornaria: I don't see a warning with gcc 14.2 (-Wall). We don't use `xdigits` as a string, only as array.

tornaria: See §6.7.8 paragraph 14 in p.126 of https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf

src/cysignals/implementation.c
cxzhong: meson directly send the parameter do not use Fortify level. So I think this part is unreachable.

src/cysignals/implementation.c
tornaria: > An array of character type may be initialized by a character string literal, optionally enclosed in braces. Successive characters of the character string literal (including the terminating null character **if there is room or if the array is of unknown size**) initialize the elements of the array.

src/cysignals/implementation.c
cxzhong: I will test this.

src/cysignals/implementation.c
cxzhong: But I see a warning in gcc 15.2. I will put the log later.

src/cysignals/implementation.c
tornaria: Please leave the test in place. This is there to catch any situation where the file is incorrectly built with fortify as shown in #226. There's no downside.

In addition, I'd suggest not doing unrelated changes in a PR, so it's easier to review. If your goal is to remove `preliminary_late_includes_cy28`, just stick to that (but also please explain why we would need / want to do that).

cxzhong: OK, I will remove unrelated part.

cxzhong: > Please leave the test in place. This is there to catch any situation where the file is incorrectly built with fortify as shown in #226. There's no downside.
> 
> In addition, I'd suggest not doing unrelated changes in a PR, so it's easier to review. If your goal is to remove `preliminary_late_includes_cy28`, just stick to that (but also please explain why we would need / want to do that).

It is in https://github.com/cython/cython/issues/2079, it is a backward compatibility setting, and the author said it will be removed in some version

src/cysignals/implementation.c
cxzhong:
```
[8/15] Compiling C object src/cys...ated_src_cysignals_signals.pyx.c.o
In file included from src/cysignals/signals.cpython-314-x86_64-linux-gnu.so.p/src/cysignals/signals.pyx.c:1133:
../../src/cysignals/implementation.c: In function ‘ulong_to_str’:
../../src/cysignals/implementation.c:163:30: warning: initializer-string for array of ‘char’ truncates NUL terminator but destination lacks ‘nonstring’ attribute (17 chars into 16 available) [-Wunterminated-string-initialization]
  163 |     const char xdigits[16] = "0123456789abcdef";
      |                              ^~~~~~~~~~~~~~~~~~
```

cxzhong: This warning, seems gcc15.2 thinks it is 17 chars

src/cysignals/implementation.c
cxzhong: the author said it is temporary and deprecated set. So I think we should not rely on this

tornaria: AFAICT the feature is not going away:
https://github.com/cython/cython/issues/2079#issuecomment-370439303

Until they decide a different way to indicate the cython compiler which files are to be late included, they'll keep the directive.

I don't see a point on making our code uglier, repeat cython code, etc.

See #49 on how this was done before the late include feature (which was added to cython by @jdemeyer for this).

PR #245, opened 2026-04-05
Fix -Wunterminated-string-initialization warning in ulong_to_str

cxzhong:
Replace string literal with explicit char array initializer to avoid the -Wunterminated-string-initialization warning from GCC15.
fix
```
[8/15] Compiling C object src/cys...ated_src_cysignals_signals.pyx.c.o
In file included from src/cysignals/signals.cpython-314-x86_64-linux-gnu.so.p/src/cysignals/signals.pyx.c:1133:
../../src/cysignals/implementation.c: In function ‘ulong_to_str’:
../../src/cysignals/implementation.c:163:30: warning: initializer-string for array of ‘char’ truncates NUL terminator but destination lacks ‘nonstring’ attribute (17 chars into 16 available) [-Wunterminated-string-initialization]
  163 |     const char xdigits[16] = "0123456789abcdef";
      |                              ^~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/cysignals/implementation.c
+++ b/src/cysignals/implementation.c
@@ -160,7 +160,7 @@ static inline void print_stderr(const char* s)
 /* str should have enough space allocated */
 static inline void ulong_to_str(unsigned long val, char *str, int base)
 {
-    const char xdigits[16] = "0123456789abcdef";
+    const char xdigits[16] = {'0','1','2','3','4','5','6','7','8','9','a','b','c','d','e','f'};
     unsigned long aux;
     int len;
 


src/cysignals/implementation.c
```diff
@@ -160,7 +160,7 @@ static inline void print_stderr(const char* s)
 /* str should have enough space allocated */
 static inline void ulong_to_str(unsigned long val, char *str, int base)
 {
-    const char xdigits[16] = "0123456789abcdef";
+    const char xdigits[16] = {'0','1','2','3','4','5','6','7','8','9','a','b','c','d','e','f'};
```
tornaria, line 163: This is ok thanks. An alternative is maybe to just change it to `const char * xdigits = "0123456789abcdef";` ?

Either way is ok, not critical, even doing an array of size 17 should be fine.

tornaria: In fact, just dropping the 16 should be ok as in
```c
    const char xdigits[] = "0123456789abcdef";
```

cxzhong: Yes, they can solve the issue both. drop 16 means let compiler to decide it.

cxzhong: But I think in the origin semantic, they set 16 means they do not want the null character?

tornaria: Yes, because it's not needed, but it doesn't hurt. This is very primitive code to avoid using `printf()` from signal handlers (printf is not reentrant, see #162 and #163) and only used for debug/warning.

PR #246, opened 2026-04-06
include Python.h first

cxzhong:
It seems CPython document said 
```
Note Since Python may define some pre-processor definitions which affect the standard headers on some systems, you must include Python.h before any standard headers are included.
It is recommended to always define PY_SSIZE_T_CLEAN before including Python.h. See [Parsing arguments and building values](https://docs.python.org/3/c-api/arg.html#arg-parsing) for a description of this macro.
```


tornaria:
I don't know what is broken without this, but ok.

cxzhong:
I also do not find something is broken, but in the document, Python.h is included first, then system headers, standard libs are included

tornaria (approved):

